PR #36, opened 2015-04-06
Using Conda to install PyQt on travis

nicoddemus:
This fixes #35.

PR #38, opened 2015-04-21
Fix segfault when app is garbage collected

jdreaver:
When I was using pytest-qt with a test suite I have, I was getting a segfault after all the tests passed. I realized that the `QApplication` instance in the `qapp` fixture was probably getting garbage collected too early. I just added a `deleteLater` call so the C++ `QApplication` object gets deleted before Python garbage collects the variable. This fixed my test suite at least, and doesn't seem to cause any harm.

### DIFF
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -383,6 +383,7 @@ def qapp():
         app = QApplication([])
         yield app
         app.exit()
+        app.deleteLater()
     else:
         yield app  # pragma: no cover
 


The-Compiler:
Oh, awesome! :smile:

I've had issues with segfaults at https://github.com/hackebrot/qutebrowser/issues/22 - so I wonder if this changes anything.

nicoddemus:
Thanks @jdreaver! :smile:

jdreaver:
No problem! 

I only encountered the segfaults after switching to PyQt5. Also, the segfault I was getting was very consistent; it happened after all tests ran every time `qtbot` was used in at least one test. 

It would be awesome if we could get more information when segfaults happen, but I think part of the nature of a segfault in Python is that Python no longer has control over the process and can't do anything else for you :/ I used tried using `gdb` once to debug a segfault in a PySide app, but there was so much non-Python noise that it didn't help.

Also, @nicoddemus, my [pyqt5 conda recipe](https://github.com/conda/conda-recipes/pull/265) is progressing well. So far, I have a recipe that builds on an old version of CentOS, and works on Arch Linux and Ubuntu 14.04. I have yet to test it on Ubuntu 12.04. Once I am confident it works, I will try to get #36 to work.

The-Compiler:
You might want to check [pytest-faulthandler](https://github.com/pytest-dev/pytest-faulthandler) respectively [faulthandler](http://faulthandler.readthedocs.org/en/latest/) in general (which is in the stdlib since Python 3.4).

It will print a python backtrace on a segfault, which sometimes is helpful, sometimes not.

Another nice trick is to prepend `catchsegv` to your commandline - not as heavy as gdb, but it'll get you a stacktrace on segfaults ([example](http://qutebrowser.org:8010/builders/ubuntu-utopic/builds/454/steps/unittests/logs/stdio)) - even when they aren't very useful sometimes, it's at least something you can search for.

jdreaver:
Wow, thanks a lot! I can't believe I've never heard of faulthandler before. That should help a lot with some more segfault heisenbugs I encounter.

As an aside, at work I happen to be making a scientific simulation GUI application using PyQt5, and we also use buildbot for continuous integration. I might have to look through qutebrowser for some design and testing ideas :smile:

The-Compiler:
Yeah, it's pretty useful! I use it to write the stacktrace to a crashlog file, and the user will see a dialog to report the crash at the next start.

Cool! If you want to talk, I suggest you join `#qutebrowser` on [Freenode](http://freenode.net/) and we can talk there instead of spamming @nicoddemus with mails :wink:

nicoddemus:
> my pyqt5 conda recipe is progressing well. So far, I have a recipe that builds on an old version of CentOS, and works on Arch Linux and Ubuntu 14.04. I have yet to test it on Ubuntu 12.04. Once I am confident it works, I will try to get #36 to work.

Thanks @jdreaver, let me know If I can help in any way. :smile:

Cheers, :beers:

nicoddemus:
Btw @jdreaver, any feedback on [pytest-faulthandler](https://github.com/pytest-dev/pytest-faulthandler) is welcome; I'm think it will not in its current state capture and display crashes if they occur after `pytest_unconfigure` is called because it's disabled at that point.

jdreaver:
Thanks! I'll check out pytest-faulthandler when I decide to tackle my segfault bugs. They aren't a high priority at the moment because they only happen in testing, and they only happen about 5% of the time.

nicoddemus:
Hey @jdreaver, I've made a few changes in the [qutecrash](https://github.com/pytest-dev/pytest-qt/compare/qutecrash) branch, and I was wondering if you can try out that branch with your test suite to see if it stops crashing? All the changes seem reasonable, but I would like first to test it against a larger test suite.

jdreaver:
I ran the branch against my PyQt5 test suite (the one that I used in this thread originally), and I saw no segfaults :smile:

However, we should probably note the backwards-incompatibility of the `QtBot` class in the changelog, since it no longer accepts an argument. I was calling this class directly because I wanted to use `qtbot` in a module-level fixture, and you can't use a function-level fixture inside a module-level fixture.

One very minor concern I have is that using a global variable to store the QApplication limits users in that they cannot have multiple QApplications behind possibly multiple different `qtbot` instances. That is, if they really want to use a different QApplication (maybe a mocked or monkeypatched one) in their own fixture with a new `QtBot` instance, they can't because all of the QtBots share a global QApplication. I don't see why anyone would do that, especially because I don't see the utility in mocking or monkeypatching a QApplication, but I just though I would point it out.

nicoddemus:
> I ran the branch against my PyQt5 test suite (the one that I used in this thread originally), and I saw no segfaults

Great! Were you getting segfaults every time your ran the test suite, or only once in a while?

Your concerns about having a global `QApplication` are valid... I much prefer to keep things explicit as in the original code, although having multiple `QApplication` (real ones, not mocked) does not seems feasible in Qt. 

Just to summarize, the main changes in the branch were:

1. Global `QApplication`, keeping a reference to it to avoid it being garbage collected;
2. Calling `deleteLater` on widgets during `QtBot` tear down;
3. Calling `processEvents` at the end of each test; 

Those were all just attempts to see if it would solve the crash on [qutebrowser](https://github.com/The-Compiler/qutebrowser)'s test suite (they didn't), but as we saw they were beneficial to your test suite as well.

Do you think we should try more testing with only changes 2 and 3 (although I see some problems implementing 3 without having a global `QApplication`), or go ahead with the changes as they are and mention them in the docs?

Many thanks for taking the time to test this!

jdreaver:
> Were you getting segfaults every time your ran the test suite, or only once in a while?

The segfault that caused me to create this PR happened consistently on teardown when I used PyQt5. There was no segfault when running your new branch with PyQt5, so I assume your new code fixes whatever the issue was too. The other segfaults I mentioned happen randomly when I run code on my CI server, but they haven't happened in a long time. I hope that switching to PyQt5 will fix them (I think a lot of them are PySide bugs).

> having multiple `QApplication` (real ones, not mocked) does not seems feasible in Qt.

That's a very good point. I think that a global `QApplication` is fine with this in mind. Furthermore, a mocked `QApplication` would probably be useless with `qtbot` anyway, since most of the functionality wouldn't work.  I'm having a really hard time thinking of a use-case for that scenario, and I think avoiding garbage collection is a much better tradeoff than supporting custom `QApplication` injection. 

If someone raises an issue later about wanting to inject a custom `QApplication` for a particular test, along with a compelling reason, we could always change it back. If someone really wants to do this, they can easily create their own test fixture for their specific need in that test, as `qtbot` probably wouldn't fit their use case anyway.

> Many thanks for taking the time to test this!

No problem! :beers: (Thanks for teaching me the beer emoji haha)

nicoddemus:
Good! Then I will merge that into master and release it in 1.4, mentioning your warnings in the backlog.

> (Thanks for teaching me the beer emoji haha)

One could argue that's the most important emoji, hehehe. :sunglasses:

PR #40, opened 2015-04-27
Capture Qt log messages

### LINKED ISSUE #39: Capturing Qt logging
Continuing the discussion from https://github.com/hackebrot/qutebrowser/issues/8#issuecomment-95246253 here :smile:

I agree it's a good idea to do a fixture similiar to `caplog`, with a `.records()` and maybe `.setLevel()` and `.atLevel()` as well.

# Levels of messages

In Qt, there are the following logging "levels"/functions - I also checked how much it's used in the whole qtbase sourcecode:

- `qFatal()`: An error which means *any* further execution of the application is useless. Calls [abort()](http://www.cplusplus.com/reference/cstdlib/abort/) immediately by default, so probably should never be captured as we'll not get to the point to display the captured output. Used 178 times, for things like when no screen was found, or you're mixing incompatible Qt libraries.
- `qCritical()`: A critical issue. Qt should IMHO use this for many things which are actually only qWarnings, but it's only used 88 times.
- `qWarning()`: Used for a lot of messages which aren't really "just warnings", but Qt is (or thinks it is) able to recover from. Used 2601 times.
- `qDebug()`: Used for debug output (1692 times), but only if you recompile Qt yourself with `-debug`.

# Handling of `qFatal()`

As said I think this should not be captured. The Qt documentation says:

> If you are using the default message handler this function will abort on Unix systems to create a core dump. On Windows, for debug builds, this function will report a _CRT_ERROR enabling you to connect a debugger to the application.

I wonder how this should be handled, as it's probably a good idea to keep that behaviour. Maybe disable the custom message handler from inside it, and then call `qFatal()` with the given message again?

# Other levels

I think `qCritical()` / `qWarning()` / `qDebug()` should all be captured and printed with failed tests by default.

However as said, I'd also like to have a way to always print `qCritical()`/`qWarning()`. Maybe some `--qt-log-always` option, and a `--qt-log-format` option so I can get it in the format I want?

nicoddemus:
WIP:

- [x] Capture and display Qt messages
- [x] Disable capture using command line 
- [x] Implement fixture to access captured messages
- [x] Control formatting using command line 
- [x] Option to output messages directly to stderr instead of capturing them (in discussion: #39)
- [x] Docs

Fixes #39


pytestqt/plugin.py
```diff
@@ -421,6 +426,81 @@ def pytest_configure(config):
...
+    Message = collections.namedtuple('Message', 'msg_type, msg')
+
+    def __init__(self):
+        self.messages = []
+
+    def handle(self, msg_type, msg):
+        """
+        Method to be installed using qInstallMsgHandler, stores each message
+        into the `messages` attribute.
+        """
```
The-Compiler: `isinstance(msg, bytes)` is usually the more idomatic way to do this :wink:

nicoddemus: Thanks, changed it.

PyQt4 in Python 3 returns `bytes`... thought odd too.

pytestqt/_tests/test_logging.py
```diff
@@ -0,0 +1,86 @@
...
+        (QtDebugMsg, 'this is a DEBUG message'),
+        (QtWarningMsg, 'this is a WARNING message'),
+        (QtCriticalMsg, 'this is a CRITICAL message'),
+    ]
+    # `messages` attribute is read-only
+    with pytest.raises(AttributeError):
+        qtlog.messages = []
+
+
+def test_fixture_with_loggin_disabled(testdir):
```
The-Compiler: loggin -> logging :wink:

pytestqt/plugin.py
```diff
@@ -421,6 +426,98 @@ def pytest_configure(config):
...
+        if isinstance(msg, bytes):
+            msg = msg.decode('utf-8', errors='replace')
+        self._messages.append(self.Message(msg_type, msg))
+
+    @property
+    def messages(self):
+        """Acess messages captured so far.
```
The-Compiler: Acess -> Access

pytestqt/plugin.py
The-Compiler: Qt4's `qInstallMsgHandler` actually takes a `QtMsgHandler` which takes a `const char *` (Qt5 takes a `QString`) - so it actually makes sense to map that to a `bytes` object in Python3.

pytestqt/_tests/test_logging.py
nicoddemus: Fixed, thanks. :smile:

pytestqt/plugin.py
nicoddemus: Fixed.

Just so you know, I was planning to ask you to review this once it was done mostly for your sake so you wouldn't need to review partial work, but if you want to review as the implementation evolves you are more than welcome. :smile:

The-Compiler: Hehe, I thought so :stuck_out_tongue: 

There are some new interesting concepts/things here for me to learn, so I thought I'll take a quick look whenever I feel like and have time, and then take a more proper look whenever you say it's done :wink:

pytestqt/plugin.py
```diff
@@ -414,6 +417,11 @@ def pytest_addoption(parser):
     parser.addini('qt_no_exception_capture',
                   'disable automatic exception capture')
 
+    parser.addoption('--no-qt-log', dest='qt_log', action='store_false',
+                     default=True)
+    parser.addoption('--qt-log-format', dest='qt_log_format',
+                     default='{rec.type_name}: {rec.message}')
```
The-Compiler: Maybe those should also be .ini options? i.e. why is `qt_no_exception_capture` a config option, but `--no-qt-log` a commandline flag?

nicoddemus: I was planning on add them as ini options too, but realized it is not really needed as the users can use `addopts` instead:

```ini
[pytest]
addopts = --qt-log-format="{rec.type_name}"
```

(I have to remember to add this to the documentation)

`qt_no_exception_capture`  is a config option because it is "part" of the configuration of the test suite, as it is not something the user should have to pass in the command line for the test suite to work. `python_files` follows the same idea, if a test suite uses `ut_` prefix for test files, it makes sense that this is configured "permanently" in the ini file, instead of having to be passed every time on the command line.

Hope my somewhat contrived explanation makes sense. :smile:

nicoddemus:
Finally merged! :sweat_smile: 

Thanks for all the support @The-Compiler! :smile:

PR #44, opened 2015-05-20
Use catchsegv for Travis

The-Compiler:
This should make it easier to find out what's happening on segfaults.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ before_script:
  - sh -e /etc/init.d/xvfb start
  - sleep 3
 
+
 script:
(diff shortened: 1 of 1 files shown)


The-Compiler:
No idea what the heck is up with the failures... Some virtualenv folder not found, and the same folder is found on the same test with another pytest version?!

nicoddemus:
Thanks! :smile: I will take a look at the other failures and see if I have any ideas.

The-Compiler:
Awesome! I'll rebase and force-push them so we see the catchsegv output.

PR #45, opened 2015-05-21
Install debugging symbols on Travis

The-Compiler:
This hopefully gives us better backtraces...

I'm not sure if all those exist - e.g. `python-pyside-dbg` will probably not be available - but let's try, and if one is missing I'll remove it.


The-Compiler:
Seems like it works - should I rebase to clean the branch up or leave it as-is?

nicoddemus:
Merged, thanks! :smile:

PR #46, opened 2015-05-21
Travis: Test with more Python versions

The-Compiler:
I just noticed 3.3 and 3.4 aren't tested - I think they should be! ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: python
 python:
   - "2.7"
   - "3.2"
+  - "3.3"
+  - "3.4"
   
 virtualenv:
   system_site_packages: true


The-Compiler:
Oh, right. If I remember correctly, Travis doesn't have system-site-packages virtualenvs for Python >3.4...

nicoddemus:
Yeah, I was trying to test at least with more recent versions in #36, but no luck yet...

The-Compiler:
I also tried to get PyQt5 working on Travis, but via the [opt-qt541](https://launchpad.net/~beineri/+archive/ubuntu/opt-qt541) ppa and building PyQt by hand (and then storing it on my server so it doesn't have to be rebuilt). Ran into some problems and then stopped caring as I have my own infrastructure which runs fine (but doesn't test PRs automatically).

I'll just wait until Travis manages to [update to Ubuntu Trusty](https://github.com/travis-ci/travis-ci/issues/2046) and then things should be much easier.

nicoddemus:
Hopefully :smile:

jdreaver:
I actually got extremely far with the PyQt5 conda recipe. It currently works well on Arch Linux and newer versions of Ubuntu, although I haven't tested it on Ubuntu 12.04. I probably spent more time than I should have already, but I would consider it *almost* done. I just have to find a little more time to tidy up :smile:

nicoddemus:
Cool! :sunglasses:

PR #43, opened 2015-05-19
[WIP] Add waitSignals.

The-Compiler:
After I was stuck with the other PR I tried adding something else I've been missing: The possibility to wait for *multiple* signals (until all have been emitted).

Here I also get a segfault:

```
pytestqt/_tests/test_wait_signal.py::test_signal_triggered[wait_function5-2000-500-False] PASSED
pytestqt/_tests/test_wait_signal.py::test_signal_triggered_multiple[wait_function0-5000-10000-20000-True] Fatal Python error: Segmentation fault

Current thread 0x00007f827e961700 (most recent call first):
  File "/home/florian/proj/pytest-qt/pytestqt/plugin.py", line 418 in wait
  File "/home/florian/proj/pytest-qt/pytestqt/_tests/test_wait_signal.py", line 32 in explicit_wait
  File "/home/florian/proj/pytest-qt/pytestqt/_tests/test_wait_signal.py", line 112 in test_signal_triggered_multiple
[...]
```

```
#0  0x0000555555d824a0 in ?? ()
#1  0x00007ffff3062921 in pyqtBoundSignal_emit (self=0x7ffff3371878, args=0x7ffff7fb8048) at ../qpy/QtCore/qpycore_pyqtboundsignal.cpp:633
#2  0x00007ffff79a9528 in PyObject_Call (func=func@entry=0x7fffe8f3fb08, arg=arg@entry=0x7ffff7fb8048, kw=<optimized out>) at Objects/abstract.c:2040
#3  0x00007ffff7a56867 in PyEval_CallObjectWithKeywords (func=0x7fffe8f3fb08, arg=0x7ffff7fb8048, kw=<optimized out>) at Python/ceval.c:4114
#4  0x00007ffff304e610 in PyQtSlot::call (this=0x555555e4b2e0, callable=0x7fffe8f3fb08, args=0x7ffff7fb8048) at ../qpy/QtCore/qpycore_pyqtslot.cpp:222
#5  0x00007ffff304e390 in PyQtSlot::invoke (this=0x555555e4b2e0, qargs=0x7fffffff73f0, self=0x0, result=0x0, no_receiver_check=false) at ../qpy/QtCore/qpycore_pyqtslot.cpp:147
#6  0x00007ffff304e0b6 in PyQtSlot::invoke (this=0x555555e4b2e0, qargs=0x7fffffff73f0, no_receiver_check=false) at ../qpy/QtCore/qpycore_pyqtslot.cpp:76
[...]

PR #42, opened 2015-05-19
Add a raising parameter to waitSignal.

The-Compiler:
I tried to add a `raising` parameter to `waitSignal` (so it'd be easier to notice if a signal wasn't emitted), but now I'm stuck.

When running the tests as-is, I get a segfault:

```
pytestqt/_tests/test_wait_signal.py::test_signal_triggered[wait_function6-2000-500-False-True] FAILED
pytestqt/_tests/test_wait_signal.py::test_signal_triggered[wait_function6-2000-500-False-True] ERROR
pytestqt/_tests/test_wait_signal.py::test_signal_triggered[wait_function7-2000-500-False-True] Fatal Python error: Segmentation fault

Current thread 0x00007f6be5e35700 (most recent call first):
  File "/home/florian/proj/pytest-qt/pytestqt/plugin.py", line 323 in wait
  File "/home/florian/proj/pytest-qt/pytestqt/plugin.py", line 357 in __exit__
  File "/home/florian/proj/pytest-qt/pytestqt/_tests/test_wait_signal.py", line 36 in context_manager_wait
  File "/home/florian/proj/pytest-qt/pytestqt/_tests/test_wait_signal.py", line 68 in test_signal_triggered
  File "/usr/lib/python3.4/site-packages/_pytest/python.py", line 204 in pytest_pyfunc_call
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 394 in execute
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 528 in _docall
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 521 in __call__
  File "/usr/lib/python3.4/site-packages/_pytest/python.py", line 1174 in runtest
  File "/usr/lib/python3.4/site-packages/_pytest/runner.py", line 90 in pytest_runtest_call
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 394 in execute
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 123 in __init__
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 107 in wrapped_call
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 393 in execute
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 528 in _docall
  File "/usr/lib/python3.4/site-packages/_pytest/core.py", line 521 in __call__
  File "/usr/lib/python3.4/site-packages/_pytest/runner.py", line 137 in <lambda>
[...]

PR #47, opened 2015-05-26
Segfault fix

The-Compiler:
This fixes the segfaults in #42 and #43.

After this is merged I'll rebase and clean up #43 first, and then #42 (to add `raising` to `waitSignals` as well).


nicoddemus:
Thanks again, all your contributions are most welcome! :smile:

The-Compiler:
I'm happy to give back - especially when it makes writing my tests easier and I benefit directly :smile:

PR #50, opened 2015-05-28
Fix building of docs on python3.

### LINKED ISSUE #49: Building docs via tox fails
From `docs/conf.py`:

```python
on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
if on_rtd:
    html_theme = 'default'
else:
    #html_theme = 'agogo'
    html_theme = 'sphinxdoc'
    # [...]
```

from `tox.ini`:

```ini
[testenv:docs]
# ...
setenv=
    READTHEDOCS=True
```

This fails when trying to build the docs locally, as it uses the `default` theme:

```
Warning, treated as error:
WARNING: 'default' html theme has been renamed to 'classic'. Please change your html_theme setting either to the new 'alabaster' default theme, or to 'classic' to keep using the old default.

ERROR: InvocationError: '/home/florian/proj/pytest-qt/.tox/docs/bin/sphinx-build -q -E -W -b html . _build'
```

I'm not sure what the best way to solve this is, as I assume `default` is valid on readthedocs and `READTHEDOCS` needs to be set so the shims are used so no Qt-backend is needed.

The-Compiler:
The Mock object which is used when building docs returned another Mock for all
attributes except some special attributes.

In Python 3, `__qualname__` (which must be a string) and `__annotations__`
(which must be a dict) got added, which caused building docs to fail.

Fixes #49.

### DIFF
--- a/pytestqt/qt_compat.py
+++ b/pytestqt/qt_compat.py
@@ -98,8 +98,10 @@ else:  # pragma: no cover
         def __getattr__(cls, name):
             if name in ('__file__', '__path__'):
                 return '/dev/null'
-            elif name == '__name__':
+            elif name in ('__name__', '__qualname__'):
                 return name
+            elif name == '__annotations__':
+                return {}
             else:
                 return Mock()
     


nicoddemus:
Nice, thanks for tracking this down.

PR #51, opened 2015-06-02
Add signal_blocker_exception test for waitSignals.

The-Compiler:
Did a look at the new tests (definitely nicer!) and thought this might be a good addition.


nicoddemus:
Certainly, thanks!

PR #57, opened 2015-06-08
Log line number fix and added better contextual support for PyQt5

nicoddemus:
@The-Compiler, what do you think? It will be a little more verbose when using PyQt5, but I think the extra information can be useful.

### DIFF
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -571,7 +571,7 @@ def pytest_addoption(parser):
                      default=True)
     if QT_API == 'pyqt5':
         default = '{rec.context.file}:{rec.context.function}:' \
-                  '{rec.context.line}: {rec.type_name}: {rec.message}'
+                  '{rec.context.line}:\n    {rec.type_name}: {rec.message}'
     else:
         default = '{rec.type_name}: {rec.message}'
     parser.addoption('--qt-log-format', dest='qt_log_format', default=default)
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -799,8 +799,9 @@ class _QtLogLevelErrorRepr(TerminalRepr):
 
     def __init__(self, item, level):
         msg = 'Failure: Qt messages with level {0} or above emitted'
-        path, line, _ = item.location
-        self.fileloc = ReprFileLocation(path, line, msg.format(level.upper()))
+        path, line_index, _ = item.location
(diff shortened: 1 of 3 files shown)


The-Compiler:
Yeah, the information definitely *is* useful!

What about adding a newline and indenting the line after the source information?

Currently:

```
/home/florian/proj/qutebrowser/git/tests/utils/test_log.py:test_unfiltered:237: QtWarningMsg: Hello World
/home/florian/proj/qutebrowser/git/tests/utils/test_log.py:test_unfiltered:238: QtWarningMsg: Another one
```

Proposed:

```
/home/florian/proj/qutebrowser/git/tests/utils/test_log.py:test_unfiltered:237:
    QtWarningMsg: Hello World
/home/florian/proj/qutebrowser/git/tests/utils/test_log.py:test_unfiltered:238:
    QtWarningMsg: Another one
```

pytestqt/plugin.py
```diff
@@ -709,7 +721,14 @@ def _handle(self, msg_type, message):
                 ignored = True
                 break
 
-        self._records.append(Record(msg_type, message, ignored))
+        if context is not None:
+            context = self._Context(
+                to_unicode(context.file),
+                to_unicode(context.function),
+                context.line,
```
The-Compiler: Note `context.file` and `context.function` can also be `None` in rare circumstances. I don't really know when it happens, but I have seen it happen (and handle it in my custom message handler).

nicoddemus:
Thanks for the suggestions, I will apply them tomorrow. :smile:

About `context.file` and `context.function` being `None`, I guess in those circumstances the output will contain `None` in those fields... not sure if there's much more that can be done since I opted to use string interpolation to configure formatting and it doesn't support conditionals.

The-Compiler:
> About context.file and context.function being None, I guess in those circumstances the output will contain None in those fields... not sure if there's much more that can be done since I opted to use string interpolation to configure formatting and it doesn't support conditionals.

That sounds good enough to me - maybe there should be a test for that scenario to make sure nothing raises `TypeError` or so because of the `None`?

nicoddemus:
> That sounds good enough to me - maybe there should be a test for that scenario to make sure nothing raises TypeError or so because of the None?

Sure, good idea.

nicoddemus:
Thanks for the review @The-Compiler! :smile:

PR #63, opened 2015-07-17
Qt model tester

The-Compiler:
There we go, a very much work-in-progress version of the model tester.

The basic tests work for PyQt5, but probably not for anything else. Some issues to look at:

# PyQt4/PySide compatibility :white_check_mark: 
- [x] How to do the `sip.cast` equivalent (so we can access private methods) in PySide?
- [x] Since the `QVariant` API changed, the tests using that need to be adjusted to work with PyQt5/PySide.
- [x] Does connecting signals like this work everywhere?

# pytest integration
- [x] What to do with the commented out `qDebug()` code? Some can probably be removed because of the assert introspection, but for some of it it might make sense to use `print` so we get that output on a failure? Or maybe introduce a new output "category" like the Qt warning messages for modeltester messages?
- [x] Could the tests (the methods in modeltester, not the tests testing modeltester) somehow be split into multiple pytest tests? Probably not as all tests are re-run when the model changes.

# Other
- [x] docs
- [x] pep8 cleanup
- [x] Expose options (e.g. for the commented out `assert self._model.data(index, QtCore.Qt.DisplayRole).isValid()` test) as arguments and/or attributes.

# Tests
- [x] unit tests with some valid and broken models
- [x] some kind of integration test based on [the PyQt5_modeltest](https://github.com/bgr/PyQt5_modeltest/blob/master/test.py) one


The-Compiler:
cc @bgr

nicoddemus:
Thanks @The-Compiler, I will take a look and probably start working on it tomorrow.

nicoddemus:
Hi @The-Compiler, only today I could take a look at the code, sorry about that.

I like the work so far, I think it would be a nice addition to pytest-qt, thanks for the initiative and having this so far down the road already! 

A few comments the points you raised:

* PyQt4/PySide compatibility

I will see if I can help in this regard, although I have little experience with the new model/view framework.

* pytest integration

I'm trying to figure out how to make this as nice as possible into pytest. I agree that we should split the model test into several sub-tests.

IMO, the ideal scenario would be that a user could just provide a new model implementation (possibly through a fixture, so it can reuse other fixtures if needed to instantiate it) and automatically pytest would "spawn" all the tests for the new model automatically. I'll have to think a little on how to do that.

The-Compiler:
Regarding pytest integration, there's another big issue in my view: The tests should be re-run automatically if the model changes, as there are also some tests checking if that happens properly.

So I should be able to set a model (which will run the tests), and then for example insert some items into the model and the tests should re-run to see the inserting worked well.

nicoddemus:
> So I should be able to set a model (which will run the tests), and then for example insert some items into the model and the tests should re-run to see the inserting worked well.

Hmm I see. But shouldn't that be part of the standard tests? I mean, we could add a new standard test which automatically changes the model (adding new items) and re-runs the tests. Of course it should be generic as to work with any model.

Aside from the issue above, I have thought about how to integrate this better in pytest and thought the following:

Create a a base class which contains all the tests, which will depend on an undeclared fixture, say **qmodel**:

```python
class BaseModelTestCollection:

    def test_basic(self, qmodel):
    def test_column_count(self, qmodel):
    def test_has_index(self, qmodel):
    def test_index(self, qmodel):
    def test_data(self, qmodel):
```

Because of how the class is named, it won't get collected by pytest. 

To test a specific implementation, a user has to subclass it and implement the fixture in the class:

```python
class TestMyModel(BaseModelTestCollection):

    @pytest.fixture
    def qmodel(self):
        return MyModelImplementation(...)
```

I think the advantage here is that one can easily add new test methods if they are specific to their class, other fixtures can be used if needed when implementing the `qmodel` fixture, and it is simple to understand.

What do you think?

nicoddemus:
Any more input here @The-Compiler? :smile:

The-Compiler:
Whoops, sorry for the delay!

> Hmm I see. But shouldn't that be part of the standard tests? I mean, we could add a new standard test which automatically changes the model (adding new items) and re-runs the tests. Of course it should be generic as to work with any model.

I'll have to think more about that - I think there could be many models which don't support inserting data from the outside. Models are basically just the data behind some kind of view.

Imagine you'd want to test a [QFileSystemModel](http://doc.qt.io/qt-5/qfilesystemmodel.html) - you can't insert/remove items as part of the test, but you might want to set a different file filter with the tester "attached" and see if the model removes/adds items correctly.

As for the pytest integration - that sounds good, but probably won't work because of the "changing model" issue.

nicoddemus:
@The-Compiler did a little more work on it:

* Created a `_debug` function that uses pytest's verbosity setting to print to stdout instead of the commented-out `qDebug` macros;
* Only left a single public function, `check` (renamed from `setup_and_run`).

After taking a closer look at the code, I decided against moving forward with my previous idea of having a base class with tests and using a fixture to provide a model, because 1) this would change the code quite a bit, and would make it harder to port changes or fixes from the original code and 2) I just realized that `run` connects itself to multiples slots which calls back to itself to check the model again, which breaks the usual pytest model.

So I'm planning to release something closer to what it was originally. We can try to come up with a solution for the "model modification" problem later, but I suspect now the user can simply call `qtmodeltester.check(model)` again after changing the model in a test.

What do you think?

The-Compiler:
Nice, thank you!

I'd leave it as-is, i.e. with the automatic rechecking when the model changes. I don't know the exact reason why the original source did it that way, but I'm guessing there was some reason for it :wink:

So what's left to do here? I think it definitely needs some more tests (I'll try to work on that this week, but no promises - I need to fix some qutebrowser bugs and release a new version).

Other than that there's some `qDebug` and `FIXME`'s left, but the rest looks good I think!

Thanks for all your work on this!

nicoddemus:
> So what's left to do here? I think it definitely needs some more tests

I was thinking of testing some (all?) Qt's builtin model classes... do you have something else in mind?

> Other than that there's some qDebug and FIXME's left, but the rest looks good I think!

I will take care of the missing `qDebug` calls, but I'm not sure I'll be able to tackle the `FIXME`, as some of them don't have any comments (I have a mind of leaving them as it is, it would make it easier to port any fixes that happen upstream, if they ever happen).

The-Compiler:
> I was thinking of testing some (all?) Qt's builtin model classes... do you have something else in mind?

These will all pass, hopefully - I'm thinking of having some custom `QAbstractItemModel`s (or `QAbstractListModel` to make things easier) which test the bad cases as well.

> I will take care of the missing `qDebug` calls, but I'm not sure I'll be able to tackle the `FIXME`, as some of them don't have any comments (I have a mind of leaving them as it is, it would make it easier to port any fixes that happen upstream, if they ever happen).

Those were added by me as I was unsure what to do with the `qDebug` stuff - now that `self._debug` is a good solution for that, I think they can go except of the "checked 10 deep" one.

While you're at it: Could you also add a link to the original implementation somewhere as a comment?

http://code.qt.io/cgit/qt/qtbase.git/tree/tests/auto/other/modeltest/modeltest.cpp

nicoddemus:
I think this is good to go... let me know if you have any more comments! 

Once this is merged I will release 1.6.0. :smile:

docs/modeltester.rst
```diff
@@ -0,0 +1,45 @@
...
+
+``pytest-qt`` includes a fixture that helps testing
+`QAbstractItemModel`_ implementations. The implementation is copied
+from the C++ code as described on the `Qt Wiki <http://wiki.qt.io/Model_Test>`_,
+and it continuously checks a model as it changes, helping to verify the state
+and catching many common errors the moment they show up.
+
+Some of the conditions caught include:
+
+* Verifying X number of rows have been inserted in the correct place after the signal rowsAboutToBeInserted() says X rows will be inserted.
```
The-Compiler: Maybe add `` around the function names here to render them monospace?

tests/test_modeltest.py
```diff
@@ -0,0 +1,49 @@
...
+    model = QFileSystemModel()
+    model.setRootPath(str(tmpdir))
+    qtmodeltester.check(model)
+    tmpdir.ensure('file3.py')
+    qtmodeltester.check(model)
+
+
+@pytest.mark.skipif(QT_API == 'pyside', reason='For some reason this fails in '
+                                               'PySide with a message about'
+                                               'columnCount being private')
```
The-Compiler: That's probably one of the reasons why the cast is needed, but there really doesn't seem any way to do it with PySide...

It seems `columnCount` is private in `QAbstractListModel` because it doesn't make any sense (it'll always be 1).

What about adding a `_column_count` method which does something like this (untested)?

```python
def _column_count(self, parent=QModelIndex()):
    if isinstance(self._orig_model, QAbstractListModel):
        assert parent is QModelIndex()
        return 1
    else:
        return self._model.columnCount(parent)
```

and then using that instead of `columnCount` everywhere?

The-Compiler:
Some additional notes apart from the comments above:

* I still think this is missing tests for *broken* models, for example to check if the modeltester fails with invalid types:

```python
@pytest.mark.parametrize('broken_role', [
    QtCore.Qt.ToolTipRole, QtCore.Qt.StatusTipRole, QtCore.Qt.WhatsThisRole,
    QtCore.Qt.SizeHintRole, QtCore.Qt.FontRole, QtCore.Qt.BackgroundColorRole,
    QtCore.Qt.TextColorRole, QtCore.Qt.TextAlignmentRole,
    QtCore.Qt.CheckStateRole
])
def test_broken_types(testdir, broken_role):
    testdir.makepyfile('''
        from pytestqt.qt_compat import QAbstractListModel, QtCore

        invalid_obj = object()  # This will fail the type check for any role

        class BrokenTypeModel(QAbstractListModel):

            def rowCount(self, parent=QtCore.QModelIndex()):
                if parent == QtCore.QModelIndex():
                    return 1
                else:
                    return 0

            def data(self, index=QtCore.QModelIndex(),
                     role=QtCore.Qt.DisplayRole):
                if role == {broken_role}:
                    return invalid_obj
                else:
                    return None

        def test_broken_type(qtmodeltester):
            model = BrokenTypeModel()
            qtmodeltester.check(model)

        def test_passing():
            # Sanity test to make sure the imports etc. are right
            pass
    '''.format(broken_role=broken_role))
    res = testdir.inline_run()
    res.assertoutcome(passed=1, failed=1)
```

(sorry, I can't easily push a commit right now)

* There's a commented out check:

```python
# While you can technically return a QVariant usually this is a
# sign of a bug in data().  Disable if this really is ok in
# your model.
# assert self._model.data(index, QtCore.Qt.DisplayRole).isValid()
```

I think this should be controlled by some documented `check_valid_display_data` attribute, and enabled by default.

pytestqt/qt_compat.py
```diff
@@ -109,16 +124,38 @@ def get_versions():
...
             return VersionTuple(qt_api_name, QtCore.PYQT_VERSION_STR,
                                 QtCore.qVersion(), QtCore.QT_VERSION_STR)
 
+        def cast(obj, typ):
+            """casts from a subclass to a parent class"""
+            return sip.cast(obj, typ)
+
+        def extract_from_variant(variant):
+            """returns python object from the given QVariant"""
+            return variant.toPyObject() if variant is not None else None
```
The-Compiler: This looks wrong - shouldn't this be in the PySide section? I don't know about PyQt4, but it's definitely the wrong thing to do for PyQt4.

I think there should be a test for this too - maybe something like this (untested):

```python
def test_qvariant(tmpdir):
    settings = QtCore.QSettings(str(tmpdir / 'foo.ini'), QSettings.IniFormat)
    if ...:
        settings.setValue("normal", QtCore.QVariant(42))
        settings.setValue("empty", QtCore.QVariant())
    else:
        settings.setValue("normal", 42)
        settings.setValue("empty", None)

    ret_normal = settings.value("normal")
    ret_empty = settings.value("empty")

    assert qt_compat.extract_from_variant(ret_normal) == 42
    assert qt_compat.extract_from_variant(ret_empty) is None
```

docs/modeltester.rst
nicoddemus: :+1: I just copied it over from Qt Wiki and thought about adding `` while doing so, but ended up forgetting to actually do it, thanks!

tests/test_modeltest.py
nicoddemus: Actually making `columnCount` private would break sub-classing contracts, as functions which receive a `QAbstractItemModel` would expect to be able to call `columnCount`, even if an instance of a `QAbstractListModel` is given.

I just noticed this in the [docs](http://doc.qt.io/qt-5/qabstractlistmodel.html) too:

> The columnCount() function is implemented for interoperability with all kinds of views, but by default informs views that the model contains only one column.

Also PySide's docs [explicitly say the same thing](https://srinikom.github.io/pyside-docs/PySide/QtCore/QAbstractListModel.html#qabstractlistmodel), so to me it seems like a bug in `PySide`. :sweat: 

I created an issue at PySide/PySide#127, let's see what they have to say.

I also tried to implement the workaround you suggested, but then I hit the same issue with `parent()` and `hasChildren()` methods, which at this point doesn't seem worthwhile to implement this workaround...

nicoddemus:
> I still think this is missing tests for broken models, for example to check if the modeltester fails with invalid types:

Added to the tests, thanks! :smile:

If you have more ideas of broken implementations that could be tested, that would be great. I have little experience with Qt's models myself.

> There's a commented out check:
> ```
> # While you can technically return a QVariant usually this is a
> # sign of a bug in data().  Disable if this really is ok in
> # your model.
> # assert self._model.data(index, QtCore.Qt.DisplayRole).isValid()
> ```
> I think this should be controlled by some documented check_valid_display_data attribute, and enabled by default.

Hmm it seems to me that the only implementation that actually returns `QVariant` objects from `data` is `PyQt4` (at least the builtin model implementations tested). IIUC, the comment is warning that usually returning a `QVariant` from `data()` is sign of a bug, so usually you should disallow this for your own models to ensure they always return only builtin types. In light of this, I have created a `data_may_return_qvariant` attribute defaulting to `False` to catch these possible bugs in user implemented models, and set that to `True` when testing the standard model types in `PyQt4`. See f60fbf9 and tell me what do you think.

nicoddemus:
> This looks wrong - shouldn't this be in the PySide section?

`PySide` doesn't even expose `QVariant` objects, it always returns builtin python types.

>  I don't know about PyQt4, but it's definitely the wrong thing to do for PyQt4.

I assume you meant `PyQt5` in the latter. Now that you mention it, I'm not sure. Weird that all tests pass... 

Oh OK, I figured what's wrong. `extract_from_variant` is used in a single place only:

```python
        # Check that the alignment is one we know about
        alignment = self._model.data(self._model.index(0, 0),
                                     QtCore.Qt.TextAlignmentRole)
        alignment = extract_from_variant(alignment)        
        if alignment is not None:
            alignment = int(alignment)
            mask = int(QtCore.Qt.AlignHorizontal_Mask |
                       QtCore.Qt.AlignVertical_Mask)
            assert alignment == alignment & mask
```

Guess what, all builtin models always returned `None` for `TextAligmentRole`, so `toPython` is never called. :sweat_smile:

nicoddemus:
Implemented a test to ensure `QVariant` related functions work across all bindings as you suggested. :sweat_smile: 

Also added a new test to cover the case of a `data()` implementation responding to a request for an `AlignmentRole`.

The-Compiler:
> Hmm it seems to me that the only implementation that actually returns `QVariant` objects from `data` is `PyQt4` (at least the builtin model implementations tested). IIUC, the comment is warning that usually returning a `QVariant` from `data()` is sign of a bug, so usually you should disallow this for your own models to ensure they always return only builtin types. In light of this, I have created a `data_may_return_qvariant` attribute defaulting to `False` to catch these possible bugs in user implemented models, and set that to `True` when testing the standard model types in `PyQt4`. See f60fbf9acf and tell me what do you think (missing docs, which I will add later if we decide the implementation is good).

I think the comment in the source is broken, and you picked it up the wrong way :wink: I didn't even notice!

In C++, of course you'd always return a `QVariant` (what else?) from `data()` - so this should say "While you can technically return an **invalid** QVariant". That's also what the code checks.

So I imagine something like this:

```python
if not self.no_display_data:
    assert self._model.data(index, QtCore.Qt.DisplayRole) != None
```

(checking with != so it works with `QVariant` as well)

The-Compiler:
> Actually making `columnCount` private would break sub-classing contracts, as functions which receive a `QAbstractItemModel` would expect to be able to call `columnCount`, even if an instance of a `QAbstractListModel` is given.
> 
> I just noticed this in the [docs](http://doc.qt.io/qt-5/qabstractlistmodel.html) too:
> 
> > The columnCount() function is implemented for interoperability with all kinds of views, but by default informs views that the model contains only one column.

Yeah, I found that weird as well - but Qt does this in `./src/corelib/itemmodels/qabstractitemmodel.h`:

```cpp
class Q_CORE_EXPORT QAbstractTableModel : public QAbstractItemModel
{
    Q_OBJECT

public:
    ...

protected:
    QAbstractTableModel(QAbstractItemModelPrivate &dd, QObject *parent);

private:
    Q_DISABLE_COPY(QAbstractTableModel)
    QModelIndex parent(const QModelIndex &child) const Q_DECL_OVERRIDE;
    bool hasChildren(const QModelIndex &parent) const Q_DECL_OVERRIDE;
};

class Q_CORE_EXPORT QAbstractListModel : public QAbstractItemModel
{
    Q_OBJECT

public:
    ...

protected:
    QAbstractListModel(QAbstractItemModelPrivate &dd, QObject *parent);

private:
    Q_DISABLE_COPY(QAbstractListModel)
    QModelIndex parent(const QModelIndex &child) const Q_DECL_OVERRIDE;
    int columnCount(const QModelIndex &parent) const Q_DECL_OVERRIDE;
    bool hasChildren(const QModelIndex &parent) const Q_DECL_OVERRIDE;
};
```

So this seems to be intended from Qt's point of view. I don't get how `QAbstractItemView` etc. work with that - I don't see any casting in there?!

> Also PySide's docs [explicitly say the same thing](https://srinikom.github.io/pyside-docs/PySide/QtCore/QAbstractListModel.html#qabstractlistmodel), so to me it seems like a bug in `PySide`. :sweat: 
> 
> I created an issue at PySide/PySide#127, let's see what they have to say.

It's definitely not a PySide bug. I see the same in PyQt when I don't `sip.cast` the model:

```
>>> from PyQt5.QtCore import QStringListModel
>>> model = QStringListModel(['a', 'b'])
>>> model.columnCount()
  File "<stdin>", line 1, in <module>
TypeError: QAbstractListModel.columnCount() is a private method
```

> I also tried to implement the workaround you suggested, but then I hit the same issue with `parent()` and `hasChildren()` methods, which at this point doesn't seem worthwhile to implement this workaround...

I still think it'd be worthwile - otherwise the model tester is essentially useless for non-tree models in PySide.

The logic would be something like this:

```python
def _column_count(self, parent=QModelIndex()):
    if isinstance(self._orig_model, QAbstractListModel) and using_pyside:
        assert parent is QModelIndex()  # maybe?
        return 1
    else:
        return self._model.columnCount(parent)

def _parent(self, index):
    if isinstance(self._orig_model, [QAbstractListModel, QAbstractTableModel]) and using_pyside:
        assert index.isValid()  # maybe?
        return QModelIndex()
    else:
        return self._model.parent(index)

def _has_children(self, parent=QModelIndex()):
    if isinstance(self._orig_model, [QAbstractListModel, QAbstractTableModel]) and using_pyside:
        assert parent is QModelIndex()  # maybe?
        return False
    else:
        return self._model.hasChildren(parent)
```

It's a bit of a hack, granted - but at least it makes the model tests work on PySide

The-Compiler:
> If you have more ideas of broken implementations that could be tested, that would be great. I have little experience with Qt's models myself.

Ideally of course we'd write broken models to test all the modeltester checks - I might contribute some, but probably not during the next few days to weeks, as I'm still struggling with qutebrowser bugs so I can finally get v0.4 out - and after that I have a backlog of things I started and never finished :wink:

nicoddemus:
> It's definitely not a PySide bug...

Argh, I should have just take a look at the header instead of believing in the docs... thanks for taking the time to point it out! :sweat: 

> I still think it'd be worthwile - otherwise the model tester is essentially useless for non-tree models in PySide.

Completely agree!

When I implemented the workaround for `columnCount`, I next stumbled on `parent`, implemented a workaround for that and then when stumbled on `hasChildren` I gave up not realizing just another workaround was all it took. :frowning: 

> Ideally of course we'd write broken models to test all the modeltester checks, I might contribute some, but probably not during the next few days to weeks

Not to worry, I will use code coverage as a guide to implement what's missing, or at least until I'm satisfied with the current code coverage.

Other than these last points, is there anything else? If not, I will try to make a `1.6.0` release later Today with the last changes discussed above, as I won't have time to work on pytest-qt the rest of the week and I would like to get a release out of the door this week if possible. :smile:

nicoddemus:
I put some more work into it, here's a quick summary:

* Implement the workarounds for private methods in PySide models, now all tests pass without needing to skip any. :smile:

* Used classes inside the test themselves and checked explicitly if `qtmodeltester.check()` would raise an error or not instead of "sub-tests" using `testdir`... this made writing and debugging the tests easier.

* Thanks for the clarification about `data_may_return_qvariant`, indeed I had completely misunderstood the comment. I changed the logic and named the attribute `data_display_may_return_none`, and also added some docs. Let me know your thoughts.

* Small code coverage fixes, like ignoring coverage for `if` statements whose sole purpose was to aid debugging, but would fail a test right after it.

* Add a specific test to cover `headerDataChanged` signal emission from models.

The last one got me thinking that the API for `qtmodeltester` is misleading. After you call `check`, `qtmodeltester` will connect and start listening for signals from the model from that point on, and when emitted they will trigger a new round of checks to ensure everything stays consistent. 

I'm thinking of changing the API a bit to convey that idea better:

```python
model = PriceListingModel()

qtmodeltester.start_checking(model)   # perform initial checks and connects to the model

# user may now change the model at will; each change may cause a test failure as the
# model is re-checked every time a signal is emitted
model.set_prices(prices)
prices.new(product=102, cost=12)

# disconnect everything; of course this is optional and is called automatically 
# by the end of the test
qtmodeltester.stop_checking(model)
```

And a context manager of course:

```python
with qtmodeltester.checking(model):
    # changes inside the block are checked
```

What do you think?

pytestqt/modeltest.py
```diff
@@ -0,0 +1,607 @@
...
+        else:
+            return self._model.parent(index)
+
+    def _has_children(self, parent=QtCore.QModelIndex()):
+        """
+        .. see:: ``_column_count``
+        """
+        model_types = (QAbstractListModel, QAbstractTableModel)
+        if isinstance(self._orig_model, model_types) and USING_PYSIDE:
+            return parent == QtCore.QModelIndex() and self._model.rowCount() > 0
```
The-Compiler: This looks wrong - it would mean any item of a `QAbstractListModel` and `QAbstractTableModel` would have children, right?

This should probably be something like the C++ source:

```cpp
bool QAbstractTableModel::hasChildren(const QModelIndex &parent) const
{
    if (parent.model() == this || !parent.isValid())
        return rowCount(parent) > 0 && columnCount(parent) > 0;
    return false;
}

bool QAbstractListModel::hasChildren(const QModelIndex &parent) const
{
    return parent.isValid() ? false : (rowCount() > 0);
}
```

I'm not sure why `QAbstractListModel::hasChildren` doesn't check the model of the index - but I think doing this is undefined behaviour anyways.

The-Compiler: Oh, it seems the Python code already matches the C++ implementation for `QAbstractListModel`.

Okay, I understand the implementation now - the *root item* has children if there are any rows.

But I guess we should also use the C++ implementation for `QAbstractTableModel::hasChildren`.

The-Compiler:
> The last one got me thinking that the API for `qtmodeltester` is misleading. After you call `check`, `qtmodeltester` will connect and start listening for signals from the model from that point on, and when emitted they will trigger a new round of checks to ensure everything stays consistent. 

That's why I originally named it `setup_and_run()` or something :wink:. Your proposal sounds good, especially with the context manager - I totally didn't think of this!

Maybe have a single-shot `check` which doesn't connect the signals as well? If the user knows they won't modify the model, then that might be the simpler API to use - "make easy things easy, and hard things possible" :wink:

The-Compiler:
Regarding releases: Maybe it's a better idea to make a 1.6.0 release without this, considering we're still reviewing and doing small incremental changes, and it isn't quite ready yet (but soon)?

I think it'd be better than trying to rush this into 1.6.0 - then we'll have the time we need.

Also, thanks again for all the great work you're doing here!

nicoddemus:
> But I guess we should also use the C++ implementation for QAbstractTableModel::hasChildren

Agree, and this made me realize there should be more tests for `QAbstractTableModel` subclasses. I will probably change the implementations to a copy of the C++ code to ensure compatibility.

> That's why I originally named it setup_and_run() or something

Yeah, sorry about that, I didn't study the implementation thoroughly before I suggesting the API. :worried: 

> Maybe have a single-shot check which doesn't connect the signals as well?

Sounds good. :smile:

> Regarding releases: Maybe it's a better idea to make a 1.6.0 release without this

I agree, I was thinking of including this into 1.6.0 because it seemed "almost ready", but I think there's more work here to do it properly. Better not rush it.

The-Compiler:
So as mentioned at the sprint, I messed up and accidentally rebased/force-pushed an older version... :cry:

@nicoddemus as soon as you're home again, can you please push your branch to another branch or repo? Then I'll take care of cleaning things up.

nicoddemus:
> So as mentioned at the sprint, I messed up and accidentally rebased/force-pushed an older version

Bad compiler! :smile:

The-Compiler:
I force-pushed again, this time based on your `modeltest-old` branch :sweat_smile: - so everything back to normal.

Changes on top of it:

- Another merge ([1](https://github.com/pytest-dev/pytest-qt/pull/63/commits/4f96e18e07cf06c26fc570a2714250ec8edf606a), [2](https://github.com/pytest-dev/pytest-qt/pull/63/commits/d3adbab5af41d56ceb586e1985a5c867c0c3bced), [3](https://github.com/pytest-dev/pytest-qt/pull/63/commits/24008bb0948a9348e7eb46ed071c402415686f57)) from master
- Added [comment](https://github.com/pytest-dev/pytest-qt/pull/63/commits/a860d4a5337a1deeff4fe8a2723010429f6d7db7) for `_cleanup` as I wondered where it was used while reading the code.
- Added some tests I worked on during the sprint ([1](https://github.com/pytest-dev/pytest-qt/pull/63/commits/777a1589211e83fc4a57931d69e52cbbd8ebe46e), [2](https://github.com/pytest-dev/pytest-qt/pull/63/commits/3881b510d2d38f5e4e0d359eda731e7c6e2b4fd7), now at 99% coverage
- [Re-added](https://github.com/pytest-dev/pytest-qt/pull/63/commits/062e3e362e3590cb08b614787d83e5aa62270c68) the `None` check in `_cleanup`
- [Removed](https://github.com/pytest-dev/pytest-qt/pull/63/commits/5561cc491d756b23056f353b866af4822b841178) casting of models as it could lead to overriden methods not actually being called - the PySide workaround is now used for PyQt too

The-Compiler:
Some tests I added failed with Qt4... After b8b3599adf440b511fa8a00d3e222fd6e8870d01 they pass with PyQt4 locally and with PySide on Travis, but not with PyQt4 on Travis... no idea what's going on there yet :worried:

The-Compiler:
My plan for tomorrow is to ~~drop Qt4 support~~ improve debug logging and see if that reveals anything useful.

If not, I'm going to try finding a machine where I can reproduce it, and then doing the same with C++ to find out if `QStandardItemModel` is just broken in some way on Qt4, or if those tests shouldn't be done at all on Qt4...

nicoddemus:
> My plan for tomorrow is to drop Qt4 support improve debug logging and see if that reveals anything useful.

OK, thanks! Didn't have much chance of working today, I plan to take a better look at this tomorrow.

> or if those tests shouldn't be done at all on Qt4...

I wouldn't discard this, actually.  

Thanks a lot for all the work Florian! :grin:

The-Compiler:
At 100% coverage now. I improved the debugging output and ended up disabling the weird parent check for Qt 4.

I'll try to add tests for the things you marked with `# pragma: no cover` before, and then I guess this is finally getting ready for another round of review :laughing: 

I wonder if we should only output the debug info with `-v` though - it doesn't seem to be too much, and it's definitely helpful if one of the related assertions fails.

nicoddemus:
LGTM, do you feel this is ready for merging?

I will later (just woke up :laughing:) work on making the lazy API and wait Until, but this needs to be merged first because it will conflict a lot with the lazy API work.

The-Compiler:
Two things are still open in my view:

- How to handle `verbose` (see https://github.com/pytest-dev/pytest-qt/pull/63#issuecomment-229327823)
- Should we change the API to be a context manager? (see https://github.com/pytest-dev/pytest-qt/pull/63#issuecomment-139099382) - on a second thought I'm not sure if we really need an API to stop checking, I don't see a reason why a test would need to do that

nicoddemus:
Oh yeah, sorry:

> I wonder if we should only output the debug info with -v though - it doesn't seem to be too much, and it's definitely helpful if one of the related assertions fails.

I agree, let's leave it as it is for now.

> Should we change the API to be a context manager?

Agree with you, don't really seem that much useful. And we can decide to add this later anyway.

I just cancelled a lot of the piled up builds on AppVeyor, let's see how this goes. :grin:

The-Compiler:
> I agree, let's leave it as it is for now.

That seems contradicting itself - currently we only display the output with `--verbose` :wink:

> Agree with you, don't really seem that much useful. And we can decide to add this later anyway.

Though if we want `check` to only check once (and not when the model is changed) like you proposed in your comment earlier, we'd need to do that now.

nicoddemus:
> That seems contradicting itself - currently we only display the output with --verbose

Oh sorry, I read that wrong and thought we were already displaying the debug output regardless of verbose. Let's go with your gut feeling then, since you have been working on this lately.

nicoddemus:
> Though if we want check to only check once (and not when the model is changed) like you proposed in your comment earlier, we'd need to do that now.

Long time since I wrote that and didn't remember exactly what the proposal was in the first place. 

My initial thought was that `check(model)` doesn't convey the idea that we are in fact checking *and* now listening for model changes and validating that as well... but `start_checking(model)` also doesn't give the idea that a check is also being done at the same time. 

I don't know, perhaps I'm overthinking this and just `check()` is fine. What do you think?

The-Compiler:
> > That seems contradicting itself - currently we only display the output with --verbose
> 
> Oh sorry, I read that wrong and thought we were already displaying the debug output regardless of verbose. Let's go with your gut feeling then, since you have been working on this lately.

I pushed a commit removing the `verbose` arg, so that's the case now.

> I don't know, perhaps I'm overthinking this and just `check()` is fine. What do you think?

That's what I think currently, yeah. As long as things are documented that way, I think it's fine.

nicoddemus:
Restarted AppVeyor build because it failed to download PyQt5 installer from source forge... :sweat: 

We probably should use `AppVeyor` cache to avoid this

The-Compiler:
Not sure if AppVeyor or sourceforge is worse :laughing:

nicoddemus:
Heh! :laughing:

nicoddemus:
After this passes do you think we can merge it?

The-Compiler:
The previous commit [passed](https://ci.appveyor.com/project/nicoddemus/pytest-qt/build/1.0.284) and in 0f3ce427929724971aed9d21348bec884ac67d5d I added `[ci skip]` as I just fixed some typos.

I still should investigate why `test_file_system_model` is hanging, but I'm not sure I'll get to that anytime soon, so I'd be okay with merging this, as it seems to work fine otherwise!

The-Compiler:
I took a quick look and noticed the filesystem model test actually runs over the root filesystem (not just `tmpdir`), for the reasons I explained in 444c95865db1b8ce6d25014019313a3439cb7c06.

I ended up just removing it. After the checks pass, I think this can finally be merged!

nicoddemus:
:tada: :boom:

The-Compiler:
Yay! :sparkles:

PR #71, opened 2015-08-04
Disconnect SignalBlocker after exiting loop.

### LINKED ISSUE #69: AttributeError: 'NoneType' object has no attribute 'signal_triggered' on teardown
I have some code which looks like this:

```python
class TestJavascriptEscape:

    def _test_escape(self, text, qtbot, webframe):
        # ...
        with qtbot.waitSignal(webframe.loadFinished, raising=True):
            webframe.setHtml(html_source)
        # ...

    @pytest.mark.parametrize('text', TESTS)
    def test_real_escape(self, webframe, qtbot, text):
        """Test javascript escaping with a real QWebPage."""
        self._test_escape(text, qtbot, webframe)

    @hypothesis.given(hypothesis.strategies.text())
    def test_real_escape_hypothesis(self, webframe, qtbot, text):
        """Test javascript escaping with a real QWebPage and hypothesis."""
        self._test_escape(text, qtbot, webframe)
```

With these fixtures:

```python
@pytest.fixture(scope='session')
def qnam():
    """Session-wide QNetworkAccessManager."""
    from PyQt5.QtNetwork import QNetworkAccessManager
    nam = QNetworkAccessManager()
    nam.setNetworkAccessible(QNetworkAccessManager.NotAccessible)
    return nam


@pytest.fixture
def webpage(qnam):
    """Get a new QWebPage object."""
    from PyQt5.QtWebKitWidgets import QWebPage

    page = QWebPage()
    page.networkAccessManager().deleteLater()
    page.setNetworkAccessManager(qnam)
    return page


@pytest.fixture
def webframe(webpage):
    """Convenience fixture to get a mainFrame of a QWebPage."""
    return webpage.mainFrame()
```

When I run this (after adding the `qtbot.waitSignal` call) on Ubuntu Trusty (and only there!) with PyQt 5.2.1, I get [this giant log](http://qutebrowser.org:8010/builders/ubuntu-trusty/builds/755/steps/unittests/logs/stdio). The gist of it is this:

```
==================================== ERRORS ====================================
____ ERROR at teardown of TestJavascriptEscape.test_real_escape_hypothesis _____

self = <_pytest.runner.SetupState object at 0x7f991bf86898>
colitem = <Function 'test_real_escape_hypothesis'>

    def _callfinalizers(self, colitem):
        # [...]

.tox/unittests/lib/python3.4/site-packages/_pytest/runner.py:356: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/unittests/lib/python3.4/site-packages/_pytest/python.py:1868: in finish
    func()
.tox/unittests/lib/python3.4/site-packages/_pytest/python.py:1826: in teardown
    next()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

qapp = <PyQt5.QtWidgets.QApplication object at 0x7f9913d87af8>
request = <SubRequest 'qtbot' for <Function 'test_real_escape_hypothesis'>>

    @pytest.yield_fixture
    def qtbot(qapp, request):
# [...]
E               Failed: Qt exceptions in virtual methods:
E               ________________________________________________________________________________
E                 File "/home/buildbotx/slaves/slave/ubuntu-trusty/build/.tox/unittests/lib/python3.4/site-packages/pytestqt/plugin.py", line 427, in _quit_loop_by_signal
E                   self.signal_triggered = True
E               
E               AttributeError: 'NoneType' object has no attribute 'signal_triggered'
E               ________________________________________________________________________________
E                 File "/home/buildbotx/slaves/slave/ubuntu-trusty/build/.tox/unittests/lib/python3.4/site-packages/pytestqt/plugin.py", line 427, in _quit_loop_by_signal
E                   self.signal_triggered = True
E               
E               AttributeError: 'NoneType' object has no attribute 'signal_triggered'

[repeated... a lot]
```

Do you have any idea what's happening there :worried:? It seems `self` is `None`... I've seen this in `QObjects` which are already `deleteLater()`'ed, but `SignalBlocker` isn't one...

I'll try to produce a minimal example later.

The-Compiler:
Otherwise, a second signal emission will still call the functions of the
already garbage-collected SignalBlocker with PyQt5 < 5.3.

The signals could already be disconnected, so the TypeError (older PyQt
versions) or RuntimeError (newer PyQt versions) which is raised in that case is
ignored.

Fixes #69.

### DIFF
--- a/pytestqt/_tests/test_wait_signal.py
+++ b/pytestqt/_tests/test_wait_signal.py
@@ -217,9 +217,8 @@ def stop_watch():
     return StopWatch()
 
 
-@pytest.mark.parametrize('multiple, raising',
-                         [(True, True), (True, False), (False, True),
-                          (False, False)])
+@pytest.mark.parametrize('multiple', [True, False])
+@pytest.mark.parametrize('raising', [True, False])
 def test_wait_signals_handles_exceptions(qtbot, multiple, raising):
     """
     Make sure waitSignal handles exceptions correctly.
@@ -241,8 +240,8 @@ def test_wait_signals_handles_exceptions(qtbot, multiple, raising):
             raise TestException
 
 
-@pytest.mark.parametrize('multiple, do_timeout',
-                         [(True, False), (True, True), (False, False), (False, True)])
+@pytest.mark.parametrize('multiple', [True, False])
+@pytest.mark.parametrize('do_timeout', [True, False])
 def test_wait_twice(qtbot, single_shot, multiple, do_timeout):
     """
     https://github.com/pytest-dev/pytest-qt/issues/69
--- a/pytestqt/_tests/test_wait_signal.py
+++ b/pytestqt/_tests/test_wait_signal.py
@@ -239,3 +239,30 @@ def test_wait_signals_handles_exceptions(qtbot, multiple, raising):
     with pytest.raises(TestException):
         with func(arg, timeout=10, raising=raising):
             raise TestException
+
+
+@pytest.mark.parametrize('multiple, do_timeout',
+                         [(True, False), (True, True), (False, False), (False, True)])
+def test_wait_twice(qtbot, single_shot, multiple, do_timeout):
+    """
+    https://github.com/pytest-dev/pytest-qt/issues/69
+    """
+    signaller = Signaller()
+
+    if multiple:
+        func = qtbot.waitSignals
+        arg = [signaller.signal]
+    else:
+        func = qtbot.waitSignal
+        arg = signaller.signal
+
+    if do_timeout:
+        with func(arg, timeout=100):
+            single_shot(signaller.signal, 200)
+        with func(arg, timeout=100):
+            single_shot(signaller.signal, 200)
+    else:
+        with func(arg):
+            signaller.signal.emit()
+        with func(arg):
+            signaller.signal.emit()
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -355,6 +355,10 @@ class _AbstractSignalBlocker(object):
         self.timeout = timeout
         self.signal_triggered = False
         self.raising = raising
+        self._timer = QtCore.QTimer()
+        self._timer.setSingleShot(True)
+        if timeout is not None:
+            self._timer.setInterval(timeout)
 
     def wait(self):
         """
@@ -368,11 +372,24 @@ class _AbstractSignalBlocker(object):
         if self.timeout is None and not self._signals:
             raise ValueError("No signals or timeout specified.")
         if self.timeout is not None:
-            QtCore.QTimer.singleShot(self.timeout, self._loop.quit)
+            self._timer.timeout.connect(self._quit_loop_by_timeout)
+            self._timer.start()
         self._loop.exec_()
         if not self.signal_triggered and self.raising:
             raise SignalTimeoutError("Didn't get signal after %sms." %
                                       self.timeout)
+
+    def _quit_loop_by_timeout(self):
+        self._loop.quit()
+        self._cleanup()
+
+    def _cleanup(self):
+        if self.timeout is not None:
+            try:
+                self._timer.timeout.disconnect(self._quit_loop_by_timeout)
+            except (TypeError, RuntimeError):
+                # already disconnected by Qt?
+                pass
 
     def __enter__(self):
         return self
@@ -426,6 +443,16 @@ class SignalBlocker(_AbstractSignalBlocker):
         """
         self.signal_triggered = True
         self._loop.quit()
+        self._cleanup()
+
+    def _cleanup(self):
+        super(SignalBlocker, self)._cleanup()
+        for signal in self._signals:
+            try:
+                signal.disconnect(self._quit_loop_by_signal)
+            except (TypeError, RuntimeError):
+                # already disconnected by Qt?
+                pass
 
 
 class MultiSignalBlocker(_AbstractSignalBlocker):
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -355,9 +355,11 @@ class _AbstractSignalBlocker(object):
         self.timeout = timeout
         self.signal_triggered = False
         self.raising = raising
-        self._timer = QtCore.QTimer()
-        self._timer.setSingleShot(True)
-        if timeout is not None:
+        if timeout is None:
+            self._timer = None
+        else:
+            self._timer = QtCore.QTimer()
+            self._timer.setSingleShot(True)
             self._timer.setInterval(timeout)
 
     def wait(self):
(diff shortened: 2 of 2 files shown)


pytestqt/plugin.py
```diff
@@ -368,12 +372,25 @@ def wait(self):
...
+            self._timer.timeout.connect(self._quit_loop_by_timeout)
+            self._timer.start()
         self._loop.exec_()
         if not self.signal_triggered and self.raising:
             raise SignalTimeoutError("Didn't get signal after %sms." %
                                       self.timeout)
 
+    def _quit_loop_by_timeout(self):
+        self._loop.quit()
+        self._cleanup()
```
nicoddemus: Shouldn't cleanup **always** be called? If so, I think you should call it as soon after `self._loop.exec_()` finishes (in line 377).

pytestqt/plugin.py
```diff
@@ -355,6 +355,10 @@ def __init__(self, timeout=1000, raising=False):
         self.timeout = timeout
         self.signal_triggered = False
         self.raising = raising
+        self._timer = QtCore.QTimer()
```
nicoddemus: Minor, but I think it would be cleaner to just create the timer if `timeout` is not `None`... then you can check for `self._timer is not None` during `_cleanup()`.

pytestqt/_tests/test_wait_signal.py
```diff
@@ -239,3 +239,30 @@ class TestException(Exception):
     with pytest.raises(TestException):
         with func(arg, timeout=10, raising=raising):
             raise TestException
+
+
+@pytest.mark.parametrize('multiple, do_timeout',
```
nicoddemus: Multiple `parametrize` marks, perhaps? :smile:

nicoddemus:
cool

@The-Compiler, could you also add a new `CHANGELOG` entry? (I just started using it instead of the [GitHub release page](https://github.com/pytest-dev/pytest-qt/releases) because it makes it easier to add entries with the PRs).

pytestqt/_tests/test_wait_signal.py
The-Compiler: I just followed the style of `test_wait_signals_handles_exceptions`, but I'll change it for both :smile:

pytestqt/plugin.py
The-Compiler: I don't really see the difference, but why not :smile:

pytestqt/plugin.py
The-Compiler: True - I first only had it in `_quit_loop_by_signal` and then noticed it's also needed on timeouts, that's why I didn't think of putting it there.

The-Compiler:
Gah... the newest commit (a4ca81a) breaks it again:

```
=================================================== ERRORS ===================================================
_____________________________ ERROR at teardown of test_wait_twice[False-False] ______________________________

self = <_pytest.runner.SetupState object at 0x7f2f7dc3ee10>
colitem = <Function 'test_wait_twice[False-False]'>

    def _callfinalizers(self, colitem):
        finalizers = self._finalizers.pop(colitem, None)
        exc = None
        while finalizers:
            fin = finalizers.pop()
            try:
>               fin()

.venv/lib/python3.4/site-packages/_pytest/runner.py:356: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.4/site-packages/_pytest/python.py:1868: in finish
    func()
.venv/lib/python3.4/site-packages/_pytest/python.py:1826: in teardown
    next()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

qapp = <PyQt5.QtWidgets.QApplication object at 0x7f2f73384168>
request = <SubRequest 'qtbot' for <Function 'test_wait_twice[False-False]'>>

    @pytest.yield_fixture
    def qtbot(qapp, request):
        """
        Fixture used to create a QtBot instance for using during testing.
    
        Make sure to call addWidget for each top-level widget you create to ensure
        that they are properly closed after the test ends.
        """
        result = QtBot()
        no_capture = _exception_capture_disabled(request.node)
        if no_capture:
            yield result  # pragma: no cover
        else:
                yield result
            if exceptions:
>               pytest.fail(format_captured_exceptions(exceptions))
E               Failed: Qt exceptions in virtual methods:
E               ________________________________________________________________________________
E                 File "/home/buildbotx/pytest-qt/pytestqt/plugin.py", line 446, in _quit_loop_by_signal
E                   self.signal_triggered = True
E               
E               AttributeError: 'NoneType' object has no attribute 'signal_triggered'
E               ________________________________________________________________________________

pytestqt/plugin.py:582: Failed
-------------------------------------------- Captured stderr call -------------------------------------------$
  File "/home/buildbotx/pytest-qt/pytestqt/plugin.py", line 446, in _quit_loop_by_signal
    self.signal_triggered = True
AttributeError: 'NoneType' object has no attribute 'signal_triggered'
===================================== 51 passed, 1 error in 6.03 seconds ====================================$
```

I'll investigate what's happening later.

The-Compiler:
Okay... I have no idea what difference the commit would make... Do you?

nicoddemus:
Hmmm if I have to guess as I can't debug this right now, I think during `_cleanup` it should disconnect also all signals that were connected (in the  `_signals` attribute). 

It's really a disappointment that signal disconnection is not handled automatically... I believe it was handled automatically in other PyQt versions (at least PyQt4).

Btw sorry if I don't answer all issues/PRs Today because I'm kinda busy, but I will probably be able to work on them tomorrow. :smile:

The-Compiler:
> Hmmm if I have to guess as I can't debug this right now, I think during `_cleanup` it should disconnect also all signals that were connected (in the `_signals` attribute).

The issue does (for a reason I don't know) not happen with MultiSignalBlocker in the first place, so I didn't want to needlessly complicate that. However, the SingleSignalBlocker test fails after changing where `_cleanup` is called... Are you okay with just reverting a4ca81a?

> It's really a disappointment that signal disconnection is not handled automatically... I believe it was handled automatically in other PyQt versions (at least PyQt4).

Yeah, as said it's also handled properly in PyQt5 5.3 and newer, this is really only an issue with ancient PyQt versions (but Ubuntu Trusty, which is still supported until 2019, comes with 5.2).

> Btw sorry if I don't answer all issues/PRs Today because I'm kinda busy, but I will probably be able to work on them tomorrow.

No worries - take your time! I just did a brain-dump of everything still in my mind :wink:

nicoddemus:
> Are you okay with just reverting a4ca81a?

Yes, we can improve that later.

> I just did a brain-dump of everything still in my mind :wink:

More than welcome to do that! :smile:

The-Compiler:
I reverted the commit and made sure the test passes on Trusty (after struggling with the VM a bit). This is good to merge from my POV now.

PR #72, opened 2015-08-04
Show Qt/PyQt/PySide versions in pytest_report_header

### LINKED ISSUE #68: Show Qt/PyQt/PySide versions in pytest_report_header
I think it'd be useful to show the Qt/PyQt/PySide versions used in the report header.

For PyQt, this would be:

- `PyQtX.QtCore.QT_VERSION_STR` - Qt version (which PyQt was compiled against)
- `PyQtX.QtCore.qVersion()` - Qt version (runtime)
- `PyQtX.QtCore.PYQT_VERSION_STR` - PyQt version

For PySide, it seems it's:

- `PySide.QtCore.__version__` - Qt version (which PySide was compiled against)
- `PySide.QtCore.qVersion()` - Qt version (runtime)
- `PySide.__version__` - PySide version

I'd submit a PR, but my TODO list gets longer and longer instead of shorter - so feel free to work on this if you feel like, otherwise it'll probably take me some weeks :laughing:

nicoddemus:
Fixes #68

### DIFF
--- a/pytestqt/_tests/test_basics.py
+++ b/pytestqt/_tests/test_basics.py
@@ -81,14 +81,21 @@ def test_widget_kept_as_weakref(qtbot):
 
 
 def test_header(testdir):
+    testdir.makeconftest(
+        '''
+        from pytestqt import qt_compat
+
+        def mock_get_versions():
+            return qt_compat.VersionTuple('PyQtAPI', '1.0', '2.5', '3.5')
+
+        assert hasattr(qt_compat, 'get_versions')
+        qt_compat.get_versions = mock_get_versions
+        '''
+    )
     res = testdir.runpytest()
-    v = get_versions()
     res.stdout.fnmatch_lines([
         '*test session starts*',
-        '{qt_api} {qt_api_version} -- Qt runtime {runtime} -- '
-        'Qt compiled {compiled}'.format(qt_api=v.qt_api,
-                                        qt_api_version=v.qt_api_version,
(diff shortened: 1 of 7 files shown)


pytestqt/_tests/test_basics.py
```diff
@@ -80,6 +80,18 @@ def test_widget_kept_as_weakref(qtbot):
...
+def test_header(testdir):
+    res = testdir.runpytest()
+    v = get_versions()
+    res.stdout.fnmatch_lines([
+        '*test session starts*',
+        '{qt_api} {qt_api_version} -- Qt runtime {runtime} -- '
+        'Qt compiled {compiled}'.format(qt_api=v.qt_api,
+                                        qt_api_version=v.qt_api_version,
+                                        runtime=v.runtime, compiled=v.compiled)
+    ])
```
The-Compiler: I'm a bit worried because the assertion basically re-runs the code under test, so even if there was something weird going on you wouldn't notice.

I don't know if it's possible/easy in conjunction with `pytester`, but maybe I'd be better to monkeypatch the version attributes and then see if the header matches a predefined output text?

nicoddemus: I had the same concern, but didn't give it much thought... shame on me! :stuck_out_tongue_winking_eye: 

Thanks for the suggestion, I will update the code once I have the change! :smile:

nicoddemus:
`testdir.runpytest` runs in a separate process, so I used a `conftest` in the test directory to do the mocking for me. :blush:

PR #73, opened 2015-08-05
Processing events before and after other fixtures are teardown

### LINKED ISSUE #67: Processing test events happens in hookwrapper
So I was looking into why the exception in #65  is actually happening for me: It seems the test finishes *after* a module I'm monkeypatching is already un-monkeypatched:

```python
@pytest.fixture(autouse=True)
def mock_modules(monkeypatch, stubs):
    monkeypatch.setattr('qutebrowser.misc.guiprocess.message',
                        stubs.MessageModule())
```

I'm guessing this is because `pytest_runtest_teardown()` is decorated with `@pytest.mark.hookwrapper` - should it really be? I think it'd make more sense to run it before other fixtures are torn down.

nicoddemus:
Fixes #67

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,13 @@
 
 # 1.5.2.dev #
 
+- `QApplication.processEvents()` is now called before and after other fixtures
+  and teardown hooks, to better try to avoid non-processed events from leaking 
+  from one test to the next. (#67, thanks @The-Compiler). 
+
 # 1.5.1 #
 
-* Exceptions are now captured also during test tear down, as delayed events will 
+- Exceptions are now captured also during test tear down, as delayed events will 
   get processed then and might raise exceptions in virtual methods; 
   this is specially problematic in `PyQt5.5`, which 
   [changed the behavior](http://pyqt.sourceforge.net/Docs/PyQt5/incompatibilities.html#pyqt-v5-5) 
--- a/pytestqt/_tests/test_basics.py
+++ b/pytestqt/_tests/test_basics.py
@@ -1,3 +1,4 @@
+import os
 import weakref
 import pytest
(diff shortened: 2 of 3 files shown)


The-Compiler:
I also pushed a test for #67 to your branch - ready to merge as far as I'm concerned!

nicoddemus:
Thanks! I tried to create a test for it, but was not satisfied what I came up with. I will change the test you slightly later to use `single_shot` fixtures (I'm a bit traumatized about `QTimer.singleShot` :stuck_out_tongue_winking_eye:) and see if I test how to test that process events were also called after other fixtures were tear down., :smile:

The-Compiler:
Let me know when the new test is ready - then I can test if it actually fails on master on my Ubuntu Trusty VM.

nicoddemus:
Will do! :smile:

nicoddemus:
Add the new test... it is a little complicated, but it covers that process events is called at the proper times. BTW, the test shows a situation where a new event might be queued during a fixture tear down, which could affect the next test. :wink:

Feel free to merge if it looks OK to you.

The-Compiler:
Nice! Just merged it - thank you! I never went that low-level with Qt's event system, but it looks good to me :smile:

The-Compiler:
By now I have several tests which fail on Windows but seem to be okay with this change - do you plan a release anytime soon? I'd appreciate it!

nicoddemus:
I was planning on releasing **1.6.0** after merging #63, which I think I can work on this week and make a release until Friday at latest. Can you wait until then? :smile:

The-Compiler:
Oh, I didn't expect you to merge this already - but sure, I can wait until then!

I'm unfortunately still busy with the next qutebrowser release, as there were some trouble with other tests on Windows and there was another bad bug found.

nicoddemus:
1.6.0 is out. :smile:

PR #78, opened 2015-08-09
Reduce verbosity in failed tests with captured exceptions

### LINKED ISSUE #77: Noisy output for exceptions in virtual methods
When there's an exception in a virtual method, e.g. with this code:

```python
from PyQt5.QtCore import QTimer

def foo():
    raise Exception


def test_foo(qtbot):
    timer = QTimer()
    timer.setSingleShot(True)
    timer.setInterval(100)
    timer.timeout.connect(foo)
    with qtbot.waitSignal(timer.timeout):
        timer.start()
```

the output is quite noisy:

```
=========================================================================== ERRORS ===========================================================================
_______________________________________________________________ ERROR at teardown of test_foo ________________________________________________________________

self = <_pytest.runner.SetupState object at 0x7ffa51cd5cf8>, colitem = <Function 'test_foo'>

    def _callfinalizers(self, colitem):
        finalizers = self._finalizers.pop(colitem, None)
        exc = None
        while finalizers:
            fin = finalizers.pop()
            try:
>               fin()

.tox/unittests/lib/python3.4/site-packages/_pytest/runner.py:356: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.tox/unittests/lib/python3.4/site-packages/_pytest/python.py:1868: in finish
    func()
.tox/unittests/lib/python3.4/site-packages/_pytest/python.py:1826: in teardown
    next()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

qapp = <PyQt5.QtWidgets.QApplication object at 0x7ffa51cca708>, request = <SubRequest 'qtbot' for <Function 'test_foo'>>

    @pytest.yield_fixture
    def qtbot(qapp, request):
        """
        Fixture used to create a QtBot instance for using during testing.
    
        Make sure to call addWidget for each top-level widget you create to ensure
        that they are properly closed after the test ends.
        """
        result = QtBot()
        no_capture = _exception_capture_disabled(request.node)
        if no_capture:
            yield result  # pragma: no cover
        else:
                yield result
            if exceptions:
>               pytest.fail(format_captured_exceptions(exceptions))
E               Failed: Qt exceptions in virtual methods:
E               ________________________________________________________________________________
E                 File "/home/florian/proj/qutebrowser/git/test_foo.py", line 4, in foo
E                   raise Exception
E               
E               Exception: 
E               ________________________________________________________________________________

.tox/unittests/lib/python3.4/site-packages/pytestqt/plugin.py:554: Failed
-------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------
  File "/home/florian/proj/qutebrowser/git/test_foo.py", line 4, in foo
    raise Exception
Exception
============================================================= 1 passed, 1 error in 0.53 seconds ==============================================================
```

- Exception text is duplicated in the captured stderr output as `sys.__excepthook__` gets called which prints in there.
- The source code of `qtbot` and internal pytest source is displayed.

Can some of this be shortened, e.g. by not calling `sys.__excepthook__`, using `pytest.fail(..., pytrace=False)` or using `__tracebackhide__` somewhere?

nicoddemus:
Fixes #77

### DIFF
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -522,13 +522,13 @@ def capture_exceptions():
 
     def hook(type_, value, tback):
         result.append((type_, value, tback))
-        sys.__excepthook__(type_, value, tback)
 
+    old_hook = sys.excepthook
     sys.excepthook = hook
     try:
         yield result
     finally:
-        sys.excepthook = sys.__excepthook__
+        sys.excepthook = old_hook
 
 
 def format_captured_exceptions(exceptions):
@@ -580,7 +580,7 @@ def qtbot(qapp, request):
         with capture_exceptions() as exceptions:
             yield result
         if exceptions:
-            pytest.fail(format_captured_exceptions(exceptions))
+            pytest.fail(format_captured_exceptions(exceptions), pytrace=False)
 
     result._close()
 
@@ -639,7 +639,8 @@ def _process_events(item):
                 app.processEvents()
             if exceptions:
                 pytest.fail('TEARDOWN ERROR: ' +
-                            format_captured_exceptions(exceptions))
+                            format_captured_exceptions(exceptions),
+                            pytrace=False)
 
 
 def pytest_configure(config):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # 1.5.2.dev #
 
+- Reduced verbosity when exceptions are captured in virtual methods
+  (#77, thanks @The-Compiler).
+
 - `QApplication.processEvents()` is now called before and after other fixtures
   and teardown hooks, to better try to avoid non-processed events from leaking 
   from one test to the next. (#67, thanks @The-Compiler). 
--- a/pytestqt/_tests/test_exceptions.py
+++ b/pytestqt/_tests/test_exceptions.py
@@ -37,6 +37,7 @@ def test_catch_exceptions_in_virtual_methods(testdir, raise_error):
             '*ValueError: mistakes were made*',
             '*1 error*',
         ])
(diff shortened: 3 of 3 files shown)


pytestqt/plugin.py
```diff
@@ -522,13 +522,13 @@ def capture_exceptions():
...
         result.append((type_, value, tback))
-        sys.__excepthook__(type_, value, tback)
 
+    old_hook = sys.excepthook
     sys.excepthook = hook
     try:
         yield result
     finally:
-        sys.excepthook = sys.__excepthook__
+        sys.excepthook = old_hook
```
The-Compiler: I think the behavior here is slightly changed - before, if there was a custom `sys.excepthook` at this point, it wasn't restored (and the system-wide one was always restored). Now the behavior is more correct.

Maybe add a test for this?

The-Compiler:
Thanks! Looks good to me apart from the comment I left.

pytestqt/plugin.py
nicoddemus: Good point, thanks! :smile:

PR #79, opened 2015-08-09
Add appveyor install scripts

### LINKED ISSUE #76: AppVeyor
Is there any reason there is no integration with [AppVeyor](http://www.appveyor.com/) yet? That would probably also be a good way to easily test PyQt5 until #36 is ready.

See the [script](https://github.com/The-Compiler/qutebrowser/blob/master/scripts/dev/ci_install.py#L51-L62) I use to install PyQt5 and my [appveyor.yml](https://github.com/The-Compiler/qutebrowser/blob/master/.appveyor.yml) - I guess something similar is also possible for PyQt4, and maybe PySide.

There's also [install.ps1](https://raw.githubusercontent.com/pypa/python-packaging-user-guide/master/source/code/install.ps1) from the Python Packaging Userguide which can be used to install various Python versions easily. (It comes with 3.4 and 2.7 only by default, I think)

nicoddemus:
As mentioned in #80:

1. I ditched the current wait signal implementation that used QEventLoop and a QTimer in favor of a busy-loop based one. This ensures we never leave the main event loop, and as bonus the code actually got simpler (0130b0f). :sweat_smile:

1. I changed the Signaller class into a fixture, which depends on the fixture that handles timers; this way we are sure Signaller instances outlive the timers, which during shutdown will disconnect from the signals too (157cf3a). Your traceback inspired me to this, as I think it is some weird shutdown order that doesn't happen everytime.

A quick review here is welcome @The-Compiler! :smile:

Fix #76 and #80

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,20 +15,10 @@ environment:
       PYTHON_ARCH: "32"
       TESTENV: "py27"
 
-    - PYTHON: "C:\\Python33"
-      PYTHON_VERSION: "3.3.x" # currently 3.3.5
-      PYTHON_ARCH: "32"
-      TESTENV: "py33"
-
     - PYTHON: "C:\\Python34"
       PYTHON_VERSION: "3.4.x" # currently 3.4.3
       PYTHON_ARCH: "32"
       TESTENV: "py34"
-
-    - PYTHON: "C:\\Python266"
-      PYTHON_VERSION: "2.6.6"
-      PYTHON_ARCH: "32"
-      TESTENV: "py26"
 
 
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,64 @@
+environment:
+  global:
+    # SDK v7.0 MSVC Express 2008's SetEnv.cmd script will fail if the
+    # /E:ON and /V:ON options are not enabled in the batch script intepreter
+    # See: http://stackoverflow.com/a/13751649/163740
+    CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\appveyor\\run_with_env.cmd"
+
+  matrix:
+
+    # Pre-installed Python versions, which Appveyor may upgrade to
+    # a later point release.
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x" # currently 2.7.9
+      PYTHON_ARCH: "32"
+      TESTENV: "py27"
+
+    - PYTHON: "C:\\Python33"
+      PYTHON_VERSION: "3.3.x" # currently 3.3.5
+      PYTHON_ARCH: "32"
+      TESTENV: "py33"
+
+    - PYTHON: "C:\\Python34"
+      PYTHON_VERSION: "3.4.x" # currently 3.4.3
+      PYTHON_ARCH: "32"
+      TESTENV: "py34"
+
+    - PYTHON: "C:\\Python266"
+      PYTHON_VERSION: "2.6.6"
+      PYTHON_ARCH: "32"
+      TESTENV: "py26"
+
+
+install:
+  - ECHO "Filesystem root:"
+  - ps: "ls \"C:/\""
+
+  - ECHO "Installed SDKs:"
+  - ps: "ls \"C:/Program Files/Microsoft SDKs/Windows\""
+
+  # Install Python (from the official .msi of http://python.org) and pip when
+  # not already installed.
+  - ps: if (-not(Test-Path($env:PYTHON))) { & appveyor\install.ps1 }
+
+  # Prepend newly installed Python to the PATH of this build (this cannot be
+  # done from inside the powershell script as it would require to restart
+  # the parent CMD process).
+  - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+
+  # Check that we have the expected version and architecture for Python
+  - "python --version"
+  - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
+
+  # Install the build dependencies of the project. If some dependencies contain
+  # compiled extensions and are not provided as pre-built wheel packages,
+  # pip will build them from source using the MSVC compiler matching the
+  # target Python version and architecture
+  - "%CMD_IN_ENV% pip install tox"
+
+build: false  # Not a C# project, build stuff at the test step instead.
+
+test_script:
+  # Build the compiled extension and run the project tests
+  - "%CMD_IN_ENV% tox -e %TESTENV%"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,6 +20,16 @@ environment:
       PYTHON_ARCH: "32"
       TESTENV: "py34"
 
+    - PYTHON: "C:\\Python33"
+      PYTHON_VERSION: "3.3.x" # currently 3.3.5
+      PYTHON_ARCH: "32"
+      TESTENV: "py33"
+
+    - PYTHON: "C:\\Python266"
+      PYTHON_VERSION: "2.6.6"
+      PYTHON_ARCH: "32"
+      TESTENV: "py26"
+
 
 install:
   - ECHO "Filesystem root:"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@
 - Reduced verbosity when exceptions are captured in virtual methods
   (#77, thanks @The-Compiler).
 
+- Internal refactoring in `waitSignal` and `waitSignals` to use a manual
+  busy loop instead of relying on `QEventLoop` and `QTimer`; depending
+  on the order objects are destroyed in tests the previous implementation
+  could leave dangling signal connections which could cause a crash (noticed
+  this on Windows with PySide and PyQt5). Should not affect tests,
+  but if users notice any behavior change between this version and `1.5.1`
+  please make sure to report it (#80, #79, thanks @The-Compiler for support).
+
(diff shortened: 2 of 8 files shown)


The-Compiler:
Unfortunately this triggers a segfault for the qutebrowser testsuite which wasn't there before :worried:

To make things worse, I can't make it happen when I only run the file it segfaults in, only when I run it together with another file...

I'm investigating and trying to minimize the example - just wanted to make sure you don't merge this yet.

The-Compiler:
This is about as minimal I can get it...

```python
import functools

import pytest
from PyQt5.QtCore import pyqtSignal, QEvent, QObject, QTimer, QEventLoop, QMetaMethod
from PyQt5.QtWidgets import QStyle, QFrame


def log_signals(obj):

    def log_slot(obj, signal, *args):
        pass

    def connect_log_slot(obj):
        metaobj = obj.metaObject()
        for i in range(metaobj.methodCount()):
            meta_method = metaobj.method(i)
            if meta_method.methodType() == QMetaMethod.Signal:
                name = bytes(meta_method.name()).decode('ascii')
                signal = getattr(obj, name)
                signal.connect(functools.partial(log_slot, obj, signal))

    connect_log_slot(obj)


class EventLoop(QEventLoop):

    def __init__(self, parent=None):
        super().__init__(parent)
        self._executing = False

    def exec_(self, flags=QEventLoop.AllEvents):
        self._executing = True
        super().exec_(flags)
        self._executing = False


def test_blah(qapp):
    obj = QObject()
    qapp.postEvent(obj, QEvent(QEvent.User))
    qapp.processEvents()


class SignalObject(QObject):

    signal = pyqtSignal()


def test_log_signals(qtbot, caplog):
    signal_obj = SignalObject()
    log_signals(signal_obj)
    with qtbot.waitSignal(signal_obj.destroyed):
        signal_obj.deleteLater()


@pytest.mark.parametrize('x', [1, 2, 3, 4, 5])
def test_foo(x):
    pass


@pytest.mark.usefixtures('qapp')
class TestEventLoop:

    def _assert_executing(self):
        assert self.loop._executing

    def test_normal_exec(self):
        self.loop = EventLoop()
        QTimer.singleShot(100, self._assert_executing)
        QTimer.singleShot(200, self.loop.quit)
        self.loop.exec_()
        assert not self.loop._executing
```

This is all copy-pasted from qutebrowser code (and test code) and then simplified - but at this point if I remove anything it doesn't segfault - and it also doesn't segfault on pytest-qt master.

It seems to segfault during `self.loop.exec_()` of `TestEventLoop.test_normal_exec`

The-Compiler:
So I could minimize it a bit more:

```python
import functools

import pytest
from PyQt5.QtCore import pyqtSignal, QEvent, QObject, QTimer, QEventLoop, QMetaMethod
from PyQt5.QtWidgets import QStyle, QFrame


class EventLoop(QEventLoop):

    def __init__(self, parent=None):
        super().__init__(parent)
        self._executing = False

    def exec_(self, flags=QEventLoop.AllEvents):
        self._executing = True
        super().exec_(flags)
        self._executing = False


def log_slot(obj):
    pass


def test_log_signals(qtbot, caplog):
    obj = QObject()
    obj.destroyed.connect(functools.partial(print, obj))
    with qtbot.waitSignal(obj.destroyed):
        obj.deleteLater()


@pytest.mark.parametrize('x', [1, 2, 3, 4, 5])
def test_foo(x):
    pass


@pytest.mark.usefixtures('qapp')
class TestEventLoop:

    def _assert_executing(self):
        assert self.loop._executing

    def test_normal_exec(self):
        self.loop = EventLoop()
        QTimer.singleShot(100, self._assert_executing)
        QTimer.singleShot(200, self.loop.quit)
        self.loop.exec_()
        assert not self.loop._executing
```

It seems to be related to connecting something to the `destroyed` object. I'm completely okay with marking this as "you just do too weird stuff, Florian" if you want - I fixed this (and the somewhat related https://github.com/The-Compiler/qutebrowser/issues/867) by not connecting to `destroyed` in `log_signals`: https://github.com/The-Compiler/qutebrowser/tree/debug-utils-tests

Though it's weird that it works before the busy-loop change... whatever. With that change, my testsuite passes with pytest-qt on the appveyor branch.

appveyor.yml
```diff
@@ -0,0 +1,64 @@
...
+      PYTHON_ARCH: "32"
+      TESTENV: "py26"
+
+
+install:
+  - ECHO "Filesystem root:"
+  - ps: "ls \"C:/\""
+
+  - ECHO "Installed SDKs:"
+  - ps: "ls \"C:/Program Files/Microsoft SDKs/Windows\""
```
The-Compiler: This seems unneeded when not installing C extensions - but it won't hurt of course.

appveyor.yml
```diff
@@ -0,0 +1,64 @@
+environment:
+  global:
+    # SDK v7.0 MSVC Express 2008's SetEnv.cmd script will fail if the
+    # /E:ON and /V:ON options are not enabled in the batch script intepreter
+    # See: http://stackoverflow.com/a/13751649/163740
+    CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\appveyor\\run_with_env.cmd"
```
The-Compiler: I think this and `run_with_env.cmd` is unneded unless building C extensions as well.

nicoddemus:
> So I could minimize it a bit more:

Thanks for minimizing it! IIUC the test you posted still crashes with the new changes, is that correct? If so, I think I will create a `acceptance_test` and but this test in there, similar to what we have on `pytest`. We can post snippets of code doing "weird stuff" as you put it to ensure it keeps working. :smile:

I think is that there is some differences in what kind of events get processed by a `QApplication.processEvents()`  and `QEventLoop.exec_()`. For example, [the docs](http://doc.qt.io/qt-4.8/qcoreapplication.html#processEvents) for `QApplication.processEvents()` mention this:

> In event you are running a local loop which calls this function continuously, without an event loop, the DeferredDelete events will not be processed.

I will think about this more carefully later, but bringing back the old implementation (based on `QEventLoop`) is still an option.

>  I fixed this (and the somewhat related The-Compiler/qutebrowser#867) by not connecting to destroyed in log_signals

The cause for the crash in this branch seems to be related to the fact that the timers were being disconnected from the `signaller` instance only after the instance was destroyed on the Python side. Changing it to make sure the timers disconnected first solved the problem. Perhaps the same idea could be applied to The-Compiler/qutebrowser#867? Perhaps a fixture that disconnects all signals connected by `log_signals`?

appveyor.yml
nicoddemus: Good point. I just copied that from `pytest's`, I think I will leave it thought as it doesn't hurt. :smile:

nicoddemus:
Hi @The-Compiler,

Oh man...

I've made a few more tests, and I really can't figure out what's going on.

I took the example from your last post and when I execute it verbatim using the code from this branch, I get a crash every time exactly like you. 

I then tied it up a bit to include into `pytest-qt`'s own test suite and out of the blue it just stopped crashing. I then backtracked and got to the point where I can make it crash just by changing the imports a bit.

With this lines:

```python
from PyQt5.QtCore import pyqtSignal, QEvent, QObject, QTimer, QEventLoop, QMetaMethod
from PyQt5.QtWidgets import QStyle, QFrame
```

I get a crash every time. If I comment those lines and add these instead:

```python
#from PyQt5.QtCore import pyqtSignal, QEvent, QObject, QTimer, QEventLoop, QMetaMethod
#from PyQt5.QtWidgets import QStyle, QFrame
from pytestqt.qt_compat import QtCore
QEventLoop = QtCore.QEventLoop
QObject = QtCore.QObject
QTimer = QtCore.QTimer
```

I get a crash **the first time** I execute it after writing the new imports, but after that **all other runs pass normally**. I tried to narrow down if it was one of the specific imports from `PyQt5` were to blame, but couldn't, the crashes come and go randomly. 

I can't figure out for the life of me what's going on. Gotta tell you, it feels like walking on glass here, where every little change might simply start crashing the whole test suite. :sweat:

nicoddemus:
A few more tests reveal that I can't reproduce the crashes in `PyQt4` or `PySide` (using the `from PyQt4.QtCore import ...` and `from PySide.QtCore import ...` lines).

nicoddemus:
All weirdness about the imports aside, if I just comment out the line that connects to the `destroyed` signal:

```python
def test_log_signals(qtbot):
    obj = QObject()
    #obj.destroyed.connect(functools.partial(print, obj))
    with qtbot.waitSignal(obj.destroyed):
        obj.deleteLater()
```

I no longer get the crashes in `PyQt5`. :sweat:

The-Compiler:
Gah, and I did just a long write-up and qutebrowser failed to get the text from the external editor (vim)... :cry:

Long story short - I suggest we don't worry about the issue. The code is doing quite special things and is maybe even just wrong.

I think what's happening is that `obj` is destroyed, emits the signal, and the slot keeps a reference using `functools.partial` to the already-destroyed object.

The part which fixed the original crash in https://github.com/The-Compiler/qutebrowser/issues/867:

```python
    with qtbot.waitSignal(signal_obj.destroyed):
        signal_obj.deleteLater()
```

won't work because:

> In event you are running a local loop which calls this function continuously, without an event loop, the DeferredDelete events will not be processed.

nicoddemus:
> Gah, and I did just a long write-up and qutebrowser failed to get the text from the external editor (vim)... :cry:

Oh my :sweat_smile: 

> Long story short - I suggest we don't worry about the issue. The code is doing quite special things and is maybe even just wrong.

OK, I agree. I will merge this later then. 

Cheers :beers:

Issue #80, opened 2015-08-10
test_wait_signal crashing on Windows


nicoddemus:
Some combination of `test_signal_triggered_multiple` parameters is crashing on Windows (see code for reference).

The-Compiler:
I thought I'll try to debug this, but it seems the tests don't run at all for me :worried:

```
C:\Users\florian\proj\pytest-qt>.\.venv\scripts\python -m py.test pytestqt\_tests -v -s
============================= test session starts =============================
platform win32 -- Python 3.4.3 -- py-1.4.29 -- pytest-2.7.2 -- C:\Users\florian\proj\pytest-qt\.venv\scripts\python.exe
PyQt5 5.4.2 -- Qt runtime 5.4.2 -- Qt compiled 5.4.2
rootdir: C:\Users\florian\proj\pytest-qt, inifile:
plugins: qt
collected 101 items

pytestqt/_tests/test_basics.py::test_basics
```

nicoddemus:
Thanks for the help! :smile:

I started getting this yesterday, will continue my investigation later today.

About your error, I think you have to set the `QT_QPA_PLATFORM_PLUGIN_PATH` variable... at least I get the same behavior if that variable is not set.

The-Compiler:
Ah, right. I have this in qutebrowser's `tox.ini` so I forgot about it as I never have to set it by hand:

```
setenv =
    QT_QPA_PLATFORM_PLUGIN_PATH={envdir}/Lib/site-packages/PyQt5/plugins/platforms
```

Will try with that set later.

The-Compiler:
I can reproduce this now - and also get a weird `AttributeError`:

```
...
pytestqt/_tests/test_wait_signal.py::test_signal_triggered_multiple[100-200-150-False-wait_function19-False] PASSED
pytestqt/_tests/test_wait_signal.py::test_signal_triggered_multiple[200-100-100-False-wait_function20-True] PASSEDAttributeError: 'Signaller' does not have a signal with the signature signal_2()

pytestqt/_tests/test_wait_signal.py::test_signal_triggered_multiple[200-100-100-False-wait_function20-True] ERROR
pytestqt/_tests/test_wait_signal.py::test_signal_triggered_multiple[200-100-100-False-wait_function21-False] PASSEDFatal Python error: Segmentation fault

Current thread 0x00000394 (most recent call first):
[...]

PR #83, opened 2015-08-30
Use QEventLoop for waitSignal implementation

### LINKED ISSUE #82: New event processing changes behaviour of tests which delete objects
In #79, you switched pytest-qt's `waitSignal` to use busy loops because using `QEventLoop` caused a weird segfault.

I recently wrote a test which fails with the new way of looping without realizing it.

The test is making sure that some function behaves properly if an object is already deleted:

```python
@pytest.mark.parametrize('delete', [True, False])
def test_set_register_stylesheet(delete, qtbot, config_stub, caplog):
    config_stub.data = {'fonts': {'foo': 'bar'}, 'colors': {}}
    obj = Obj("{{ font['foo'] }}")

    with caplog.atLevel(9):  # VDEBUG
        style.set_register_stylesheet(obj)

    records = caplog.records()
    assert len(records) == 1
    assert records[0].message == 'stylesheet for Obj: font: bar;'

    assert obj.rendered_stylesheet == 'font: bar;'

    if delete:
        with qtbot.waitSignal(obj.destroyed):
            obj.deleteLater()

    config_stub.data = {'fonts': {'foo': 'baz'}, 'colors': {}}
    style.get_stylesheet.cache_clear()
    config_stub.changed.emit('fonts', 'foo')

    if delete:
        expected = 'font: bar;'
    else:
        expected = 'font: baz;'
    assert obj.rendered_stylesheet == expected
```

The code under test:

```python
def set_register_stylesheet(obj):
    """Set the stylesheet for an object based on it's STYLESHEET attribute.

    Also, register an update when the config is changed.

    Args:
        obj: The object to set the stylesheet for and register.
             Must have a STYLESHEET attribute.
    """
    qss = get_stylesheet(obj.STYLESHEET)
    log.config.vdebug("stylesheet for {}: {}".format(
        obj.__class__.__name__, qss))
    obj.setStyleSheet(qss)
    objreg.get('config').changed.connect(
        functools.partial(update_stylesheet, obj))


def update_stylesheet(obj):
    """Update the stylesheet for obj."""
    if not sip.isdeleted(obj):
        obj.setStyleSheet(get_stylesheet(obj.STYLESHEET))
```

Those tests now fail, presumably because the object never got deleted:

test_set_register_stylesheet[True]:

```
>       assert obj.rendered_stylesheet == expected
E       assert 'font: baz;' == 'font: bar;'
E         - font: baz;
E         ?         ^
E         + font: bar;
E         ?         ^
```

test_set_register_stylesheet[False]:

```
    @pytest.mark.parametrize('delete', [True, False])
    def test_set_register_stylesheet(delete, qtbot, config_stub, caplog):
        config_stub.data = {'fonts': {'foo': 'bar'}, 'colors': {}}
        obj = Obj("{{ font['foo'] }}")
    
        with caplog.atLevel(9):  # VDEBUG
            style.set_register_stylesheet(obj)
    
        records = caplog.records()
        assert len(records) == 1
>       assert records[0].message == 'stylesheet for Obj: font: bar;'
E       assert 'stylesheet f...j: font: baz;' == 'stylesheet fo...j: font: bar;'
E         - stylesheet for Obj: font: baz;
E         ?                             ^
E         + stylesheet for Obj: font: bar;
E         ?                             ^
```

Here I can imagine I'm not the only one with such a kind of test...

I think the best way to implement SignalBlocker would actually be a [QSignalSpy](http://doc.qt.io/qt-5/qsignalspy.html) (which would also make #64 easier) - however it only takes one signal, so `SignalBlocker.connect` and `MultiSignalBlocker` won't work with that...

nicoddemus:
Could you check that this change now works for you @The-Compiler? 

Fix #82

### DIFF
--- a/pytestqt/_tests/test_wait_signal.py
+++ b/pytestqt/_tests/test_wait_signal.py
@@ -3,7 +3,7 @@ import time
 
 import pytest
 
-from pytestqt.qt_compat import QtCore, Signal
+from pytestqt.qt_compat import QtCore, Signal, QT_API
 
 
 def test_signal_blocker_exception(qtbot):
@@ -283,3 +283,21 @@ def test_wait_twice(qtbot, timer, multiple, do_timeout, signaller):
             signaller.signal.emit()
         with func(arg):
             signaller.signal.emit()
+
+
+@pytest.mark.skipif(QT_API == 'pyside', reason='test crashes PySide')
+def test_destroyed(qtbot):
+    """Test that waitSignal works with the destroyed signal (#82).
+
+    For some reason, this crashes PySide although it seems perfectly fine code.
+    """
+    import sip
+
+    class Obj(QtCore.QObject):
+        pass
+
+    obj = Obj()
+    with qtbot.waitSignal(obj.destroyed):
+        obj.deleteLater()
+
+    assert sip.isdeleted(obj)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,6 @@
 - Reduced verbosity when exceptions are captured in virtual methods
   (#77, thanks @The-Compiler).
 
-- Internal refactoring in `waitSignal` and `waitSignals` to use a manual
-  busy loop instead of relying on `QEventLoop` and `QTimer`; depending
-  on the order objects are destroyed in tests the previous implementation
-  could leave dangling signal connections which could cause a crash (noticed
-  this on Windows with PySide and PyQt5). Should not affect tests,
-  but if users notice any behavior change between this version and `1.5.1`
-  please make sure to report it (#80, #79, thanks @The-Compiler for support).
-
 - `QApplication.processEvents()` is now called before and after other fixtures
   and teardown hooks, to better try to avoid non-processed events from leaking 
   from one test to the next. (#67, thanks @The-Compiler). 
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -351,9 +351,16 @@ class _AbstractSignalBlocker(object):
     """
 
     def __init__(self, timeout=1000, raising=False):
+        self._loop = QtCore.QEventLoop()
         self.timeout = timeout
         self.signal_triggered = False
         self.raising = raising
+        if timeout is None:
+            self._timer = None
+        else:
+            self._timer = QtCore.QTimer(self._loop)
+            self._timer.setSingleShot(True)
+            self._timer.setInterval(timeout)
 
     def wait(self):
         """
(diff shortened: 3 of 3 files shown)


pytestqt/_tests/test_wait_signal.py
```diff
@@ -283,3 +283,21 @@ def test_wait_twice(qtbot, timer, multiple, do_timeout, signaller):
             signaller.signal.emit()
         with func(arg):
             signaller.signal.emit()
+
+
+@pytest.mark.skipif(QT_API == 'pyside', reason='test crashes PySide')
+def test_destroyed(qtbot):
+    """Test that waitSignal works with the destroyed signal (#82).
+
+    For some reason, this crashes PySide although it seems perfectly fine code.
```
The-Compiler: I'm guessing `import sip` and `sip.isdeleted` won't work on PySide though - or did you test it before adding that?

The-Compiler:
Yep - the qutebrowser testsuite now works fine with this branch!

pytestqt/_tests/test_wait_signal.py
nicoddemus: This test only runs for PyQt4 and PyQt5... note the `pytest.mark.skipif` at the top. :wink:

nicoddemus:
If you don't have anything else to comment here @The-Compiler, feel free to merge please. :smile:

pytestqt/_tests/test_wait_signal.py
The-Compiler: Sorry - I was trying to say "maybe it segfaulted with PySide because `sip.isdeleted` raised an exception and that caused a segfault for some reason" - but I guess it already segfaulted *before* you added `sip.isdeleted` :wink:

nicoddemus: Oh yes, it crashed inside `QEventLoop.exec()` method... decided to don't worry about it too much.

The-Compiler:
It seems the skipped test from #80 isn't contained here, and it still passes on AppVeyor, right? Did you find a way to fix that, or is it just luck?

nicoddemus:
I think the main thing which fixed those crashes was this point:

>> I changed the Signaller class into a fixture, which depends on the fixture that handles timers; this  way we are sure Signaller instances outlive the timers, which during shutdown will disconnect from the signals too (157cf3a). Your traceback inspired me to this, as I think it is some weird shutdown order that doesn't happen everytime.

I ended up ditching the code that used `QEventLoop` at that time because it got simpler to understand IMO.

The-Compiler:
Ah, I see - I thought ditching `QEventLoop` caused the crash to disappear, but so it wasn't that - merging!

PR #84, opened 2015-08-31
Add testing environments for PyQt4 and PyQt5

### LINKED ISSUE #75: Providing pyqt testenvs in tox.ini
What about adjusting `tox.ini` so it has test environments for every Python version for `-pyside`, `-pyqt4` and `-pyqt5`, which then set `PYTEST_QT_API` correctly? For PyQt, those would probably simply use `system_site_packages = True` and rely on it being installed system-wide, for the `-pyside` envs it would install it from PyPI.

The default environment list would then be something like `py{26,27,32,33,34}-{pyside,pyqt4,pyqt5}, docs` (if I understood tox' [generative envlist](https://testrun.org/tox/latest/config.html#generative-envlist) feature correctly).

After that, we could probably adjust `.travis.yml` to use tox as well, similar to how [pytest does it](https://github.com/pytest-dev/pytest/blob/master/.travis.yml), right?

nicoddemus:
Fix #75


tox.ini
```diff
@@ -1,13 +1,22 @@
...
-    pyside    
+deps=pytest
+    py{26,27,32,33,34}-pyside: pyside
 commands=py.test {posargs}
+setenv=
+    py{26,27,32,33,34}-pyside: PYTEST_QT_API = pyside
+    py{26,27,32,33,34}-pyqt4: PYTEST_QT_API = pyqt4
+    py{26,27,32,33,34}-pyqt5: PYTEST_QT_API = pyqt5
+sitepackages=True
+passenv=QT_QPA_PLATFORM_PLUGIN_PATH
```
The-Compiler: Why not use `setenv` for `QT_QPA_PLATFORM_PLUGIN_PATH` instead of telling the user to do it?

That's what I have in qutebrowser's `tox.ini`:

```
# https://bitbucket.org/hpk42/tox/issue/246/ - only needed for Windows though
setenv =
    QT_QPA_PLATFORM_PLUGIN_PATH={envdir}/Lib/site-packages/PyQt5/plugins/platforms
```

The-Compiler: Adding `DISPLAY` and `XAUTHORITY` to `passenv` will be needed on Linux.

tox.ini
```diff
@@ -1,13 +1,22 @@
...
-envlist = py26, py27, py32, py33, py34, docs
+envlist = py{27,34}-pyqt4, py34-pyqt5, py{26,27,32,33,34}-pyside, docs
+
+
 [testenv]
-deps=pytest 
-    pyside    
+deps=pytest
+    py{26,27,32,33,34}-pyside: pyside
 commands=py.test {posargs}
```
The-Compiler: This actually runs the system-wide pytest for me - I think `{envpython} -m py.test` would be better.

tox.ini
nicoddemus: About `QT_QPA_PLATFORM_PLUGIN_PATH`, I don't set it because it is expected to be installed in system python, not in the virtual environment, and I couldn't figure how to access the system python from `tox`. `qutebrowser` sets it because it copies the files over to the virtual environment, right? Perhaps that's what should be done here as well.

The-Compiler: Yep, qutebrowser symlinks (on Linux/OS X) or copies (on Windows) PyQt into the virtualenv as I wanted a clearly isolated environment. Feel free to copy `link_pyqt.py` into your repo and license it as LGPL3 if you want to do the same.

nicoddemus:
All done, many thanks for the suggestions! :smile:

Could you take a look at `link_pyqt.py`, and see if the new license text is appropriate? I have little experience in this regard, so let me know if anything is missing.

The-Compiler:
Looks good to me! So can this be merged? It works fine for me, and definitely makes contributing a lot easier when you don't have to remember to set things up correctly every time :smile:

nicoddemus:
Definitely! :smile:

PR #91, opened 2015-09-02
Introduce backward compatibility for symbols previously available in pytestqt.plugin

### LINKED ISSUE #90: Plugin move broke using pytestqt.plugin.SignalBlocker
I have a test where I use a `pytestqt.plugin.SignalBlocker` manually, because I need it in a session-scoped fixture, and `qtbot` is a function-scoped one.

This broke with #87 - I think every public class/function which was in `plugin.py` should also be imported there after the move. Who knows what other crazy stuff people did :wink:

nicoddemus:
Fix #90

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # 1.6.0.dev #
 
-# 1.5.2.dev #
-
 - Reduced verbosity when exceptions are captured in virtual methods
   (#77, thanks @The-Compiler).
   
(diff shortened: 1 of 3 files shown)


The-Compiler:
The tests look good and my testsuite works again - I'll merge as soon as the AppVeyor check is done (even though I don't expect platform differences - but better safe than sorry).

nicoddemus:
:+1: Thanks! :smile:

PR #93, opened 2015-09-15
Add support for PyQt4 v2 API

montefra:
PyQt4 offers for python2 [two incompatible API](http://pyqt.sourceforge.net/Docs/PyQt4/incompatible_apis.html). The API must be chosen before importing anything that uses it, which apparently pytest-qt does.
The default API for PyQt4 is v1 and this makes impossible to test PyQt4 applications using the second version.

This PR aims at allowing to set the api version to 2 before importing PyQt4 setting the environment variable ``PYTEST_QT_API`` to ``pyqt4v2``

For python3 the two apis should be identical.

Using the code in this PR I managed to run my tests both in python 2 and python 3 as:

    PYTEST_QT_API=pyqt4v2 py.test

I've run ``tox``, but it got stuck at pyside installation. All the tests with PyQt* when through, with only one failure due to a ``assert 0``. [Here](https://gist.github.com/montefra/878bfd839400e426b788) is the log of the run


scripts/install-qt.py
```diff
@@ -57,8 +57,11 @@ def apt_get_install(packages):
         subprocess.check_call(['sudo', 'apt-get', 'install', '-qq'] + packages)
 
     py3k = sys.version_info[0] == 3
-    if os.environ['PYTEST_QT_API'] in ('pyqt4', 'pyqt5'):
-        pyqt_ver = os.environ['PYTEST_QT_API'][-1]
+    if os.environ['PYTEST_QT_API'] in ('pyqt4', 'pyqt4v2', 'pyqt5'):
+        if os.environ['PYTEST_QT_API'] in ('pyqt4', 'pyqt5'):
+            pyqt_ver = os.environ['PYTEST_QT_API'][-1]
+        else:
+            pyqt_ver = os.environ['PYTEST_QT_API'][-3]
```
The-Compiler: At this point I think it's easier to just do a dict:

```python
pyqt_versions = {
    'pyqt4': '4',
    'pyqt4v2': '4',
    'pyqt5': '5',
}

if os.environ['PYTEST_QT_API'] in pyqt_versions:
    pyqt_ver = pyqt_versions[os.environ['PYTEST_QT_API']]
```

(untested)

The-Compiler:
The failing test is actually not becase of the `assert 0` - it's running pytest in a subprocess and checking the output, and not finding something it expects.

#63 will probably need some adjustments after this as well.

scripts/install-qt.py
montefra: I like this: it's much clearer. I'll modify my changes.

montefra:
Now when the ``PYTEST_QT_API=pyqt4v2`` is selected all the possible options are switched to api v2.

@The-Compiler: do we want to give the possibility to chose which objects use v2?

If yes we could add a command line options and(/or) configuration option like ``pyqt4v2_objects``:

* if not given, or ``pyqt4v2_objects = all``, switch all objects to v2
* otherwise accepts a comma (or space) separated list of objects that must be switched to v2; e.g.:

        pyqt4v2_objects = QString, QDateTime

The-Compiler:
I don't really have an opinion on that - I only use PyQt5 myself.

Do you think people would need it? If not, I'd say let's avoid the unnecessary complexity unless someone requests it - but in the end, @nicoddemus will have to decide :wink:

montefra:
> Do you think people would need it?

You never know.

> If not, I'd say let's avoid the unnecessary complexity unless someone requests it - but in the end, @nicoddemus will have to decide ￼

That sounds fine: less work for me :+1:

nicoddemus:
>  do we want to give the possibility to chose which objects use v2?

Let's wait for someone to request that. :smile:

Thanks a lot for the contribution @montefra!

montefra:
@nicoddemus : a pleasure. Since I'm using this feature in one of my applications: when approximately is it going to land on PyPI and with which version?

nicoddemus:
This would be 1.7.0... I will make a new release shortly. :smile:

montefra:
great. Thanks

nicoddemus:
Released `1.7.0`. :smile:

Thanks again!

montefra:
Is there a way to set ``PYTEST_QT_API`` via command line or configuration file?

The-Compiler:
It's an environment variable, so you'd do something like

`PYTEST_QT_API=pyqt5 py.test ...`

Or use [tox](http://tox.readthedocs.org/en/latest/index.html) and set it there.

montefra:
Thanks.

It's probably time for me to start using tox

Issue #70, opened 2015-08-04
test_context_none fails


The-Compiler:
Sorry for all the reports! :laughing:

Getting this when running the tests:

```
============================================ test session starts =============================================
platform linux -- Python 3.4.0 -- py-1.4.30 -- pytest-2.7.2
qt-api: pyqt5
rootdir: /home/buildbotx/pytest-qt, inifile: 
plugins: qt
collected 24 items 

pytestqt/_tests/test_logging.py .......................F

================================================== FAILURES ==================================================
_____________________________________________ test_context_none ______________________________________________

testdir = <TmpTestdir local('/tmp/pytest-24/testdir/test_context_none0')>

    @pytest.mark.skipif(QT_API != 'pyqt5',
                        reason='Context information only available in PyQt5')
    def test_context_none(testdir):
        """
        Sometimes PyQt5 will emit a context with some/all attributes set as None
        instead of appropriate file, function and line number.
    
        Test that when this happens the plugin doesn't break.
    
        :type testdir: _pytest.pytester.TmpTestdir
        """
        testdir.makepyfile(
            """
            from pytestqt.qt_compat import QtWarningMsg
    
            def test_foo(request):
                log_capture = request.node.qt_log_capture
                context = log_capture._Context(None, None, None)
                log_capture._handle_with_context(QtWarningMsg,
                                                 context, "WARNING message")
                assert 0
            """
        )
        res = testdir.runpytest()
        res.stdout.fnmatch_lines([
            '*Failure*',
>           '*None:None:None:*',
        ])
E       Failed: remains unmatched: '*Failure*', see stderr

/home/buildbotx/pytest-qt/pytestqt/_tests/test_logging.py:417: Failed
-------------------------------------------- Captured stdout call -------------------------------------------$
running ['/home/buildbotx/pytest-qt/.venv/bin/python3', '/home/buildbotx/pytest-qt/.venv/lib/python3.4/site-p$
ckages/pytest.py', '--basetemp=/tmp/pytest-24/testdir/test_context_none0/runpytest-0'] curdir= /tmp/pytest-24$
testdir/test_context_none0
============================= test session starts ==============================
platform linux -- Python 3.4.0 -- py-1.4.30 -- pytest-2.7.2
qt-api: pyqt5
rootdir: /tmp/pytest-24/testdir/test_context_none0, inifile: 
plugins: qt
collected 1 items

test_context_none.py F

=================================== FAILURES ===================================
[...]

PR #96, opened 2015-09-21
Adjust supported python versions.

The-Compiler:
This tests pytest-qt with all supported Python versions on Travis, and drops
support for Python < 3.3.

3.0 and 3.1 were never tested anyways, and coverage.py 4.0 dropped support for
3.2 so the tests started failing on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ env:
  - PYTEST_VERSION=2.7.2 PYTEST_QT_API=pyqt5
 
 install:
- - sudo add-apt-repository --yes ppa:ubuntu-sdk-team/ppa
  - sudo apt-get update
 
  # Xvfb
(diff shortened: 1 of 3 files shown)


nicoddemus:
Thanks! I did a quick try to fix this in the [pytest2.8](https://github.com/pytest-dev/pytest-qt/tree/pytest2.8) branch, thinking it might be an incompatibility between pytest-2.8 and pytest-cov but then noticed that the failure is related with unsupported `u""` string in `3.2`.

I did not know Travis had updated its supported versions, thanks for letting me know. I wonder if they are still using an ubuntu version which only has pre-built binaries for 3.2?

The-Compiler:
I guess that's the case, yeah...

I think it works when not using `system_site_packages: true`, but that doesn't really work for pytest-qt... I'll think of a good solution later.

The-Compiler:
After playing with using `tox` on Travis and hitting some issues, I decided to use the `services: docker` hack to get an Ubuntu Trusty on Travis.

This means the tests are now running with 3.4 instead of 3.2, and seem to pass nicely again. In addition, we can test PyQt5 with Python 3 now, but unfortunately it fails because of #70, so I left it as expected failure for now.

As a downside, 2.6 and 3.3 aren't tested (well, 2.6 never was). Maybe those should be tested on AppVeyor?

AppVeyor still has some backlog from my experiments and doesn't seem to use the pytestbot account, so it'd be nice if you could cancel the older builds there. Travis is actually finished and not marked as finished here...

nicoddemus:
Thanks a lot Florian! :smile:

I cancelled the older builds, will merge once it passes. I created #101 about testing py26 on AppVeyor, and will take a look at #70 once I have some time later this week.

PR #100, opened 2015-10-12
Extend log patterns

The-Compiler:
There are actually a few unrelated changes in here, but I hope you don't mind (at least they have their own commits :wink:).

* Fix test_logging_fails_ignore_mark.
* Add extend keyword argument to qt_log_ignore mark.
* Validate qt_log_ignore keyword arguments.
* Remove unneeded imports.
* Add .cache and .venv to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ __pycache__
 /.tox
 .env*
 .coverage
+/.cache
+/.venv
--- a/pytestqt/logging.py
+++ b/pytestqt/logging.py
@@ -24,7 +24,11 @@ class QtLoggingPlugin(object):
             return
         m = item.get_marker('qt_log_ignore')
         if m:
-            ignore_regexes = m.args
+            if m.kwargs.get('extend', False):
+                config_regexes = self.config.getini('qt_log_ignore')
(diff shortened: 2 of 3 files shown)


nicoddemus:
I don't mind at all, thanks for the PR Florian! :smile:

I'm investigating the errors on py27 on Windows.

The-Compiler:
The ones on Linux are because coverage 4.0 doesn't support Python 3.2 anymore - maybe it's time to drop 3.2 support too?

nicoddemus:
Yeah, the only problem is that `pytest-qt` is available on `3.2` only on the Ubuntu version supported by Travis... I have to find some time and investigate how to get around that. :expressionless:

The-Compiler:
Oh, right, reminds me of #96 - one possibility would probably be to use tox on Travis too, with the `link_pyqt` script. I hope I get time to take a look later today.

nicoddemus:
The failures on Windows are not related to this change... I will have to fix that on `pytest` itself.

nicoddemus:
Opened pytest-dev/pytest#1133 to discuss how to properly solve this.

Issue #107, opened 2015-10-27
Add qtbot.wait()


rossant:
Would it be possible to add a simple `qtbot.wait()` method to wait for X milliseconds? For example, the following seems to work currently:

```python
def wait(x):
    from pytestqt.qt_compat import QtTest
    QtTest.QTest.qWait(x)
```

The-Compiler:
I just wanted to comment here, but I'll do in #31 then :wink:

nicoddemus:
Thanks @rossant. It was not exposed because PySide does not support it, but it is fine to only expose this methods to PyQt.

@The-Compiler thanks! :smile:

PR #108, opened 2015-10-27
Setup/teardown exception capturing as early/late as possible to catch all exceptions

### LINKED ISSUE #105: Setup/teardown exception capturing as early/late as possible to catch all exceptions
I noticed sometimes, an exception happens and PyQt5 calls `abort`, even with `pytest-qt`.

It seems this happens when an exception happens in a virtual Qt method during setup or teardown, e.g. with a `yield_fixture` or when `qtbot` calls `close()` on the widget.

I think `pytest-qt` should do the exception capturing setup/teardown as early/late as possible.

See https://github.com/The-Compiler/qutebrowser/pull/1048#issuecomment-150660769 for an example.

### LINKED ISSUE #106: Close widgets before tearing down other fixtures
In https://github.com/The-Compiler/qutebrowser/pull/1048#issuecomment-150660769 I noticed a problem: I use a fixture to provide a `config` object in my object registry (basically a dict mapping strings to global objects).

However, on OS X, closing the widget seems to call `minimumSizeHint` on it, which causes the widget to get the config - but at the point `qtbot` calls `close` on the widget, the other fixture already is torn down.

I think the widget cleanup from `qtbot` should happen as early as possible.

nicoddemus:
Now exception capturing is setup as early as possible, and disabled as late as possible. Also, this new approach allows us to distinguish exceptions captured during setup/call/teardown, which provides a more friendly message when the exception occurred inside the test call (as opposed during `qtbot`'s teardown).

Fixes #105
Fixes #106

### DIFF
--- a/pytestqt/exceptions.py
+++ b/pytestqt/exceptions.py
@@ -1,6 +1,7 @@
 from contextlib import contextmanager
 import sys
 import traceback
+import pytest
 
 
 @contextmanager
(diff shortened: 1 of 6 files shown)


pytestqt/exceptions.py
```diff
@@ -10,17 +11,54 @@ def capture_exceptions():
...
+
+    def fail_if_exceptions_occurred(self, when):
+        """calls pytest.fail() with an informative message if exceptions
+        have been captured so far. Before pytest.fail() is called, also
+        finish capturing.
+        """
+        if self.exceptions:
+            self.finish()
+            exceptions = self.exceptions[:]
+            self.exceptions[:] = []
```
The-Compiler: Why not simply:

```python
exceptions = self.exceptions
self.exceptions = []
```

here?

pytestqt/plugin.py
```diff
@@ -90,15 +83,45 @@ def pytest_addoption(parser):
                          'default: "{0}"'.format(default))
 
 
-@pytest.mark.hookwrapper
+@pytest.hookimpl(hookwrapper=True, tryfirst=True)
```
The-Compiler: I think `@pytest.hookimpl` is a pytest 2.8 thing - not sure if we should already break 2.7 compatibility as there are still some people with 2.8 regressions I think.

Maybe use `@pytest.mark.hookimpl` for now if `@pytest.hookimpl` is unavailable?

The-Compiler:
Looks great apart from the things I commented on! It doesn't fix #106 though - I added a comment there to clear things up hopefully.

With this, I can finally develop complex fixtures without getting an `abort()` thrown at me :wink:

pytestqt/exceptions.py
nicoddemus: Of course! :sweat_smile: 

Done

pytestqt/plugin.py
nicoddemus: Good point, I will use only `pytest.mark.*` variants to keep 2.7 compatibility.

nicoddemus:
Implemented the changes we talked about. Could you please test the new commits again? Thanks!

The-Compiler:
Yay, thanks again! I'll test and merge this either later today or tomorrow.

nicoddemus:
Nice, thanks!

pytestqt/plugin.py
```diff
@@ -90,32 +83,58 @@ def pytest_addoption(parser):
                          'default: "{0}"'.format(default))
 
 
+@pytest.hookimpl(hookwrapper=True, tryfirst=True)
```
The-Compiler: This is also 2.8 specific :wink:

nicoddemus: Ooops. :sweat_smile:

The-Compiler:
Looks good now, and fixes the issues I've had :+1:

I think I'll give #107 a quick try on my way home (unless you want to), and after that, what do you think about a release?

nicoddemus:
> I think I'll give #107 a quick try on my way home (unless you want to), and after that, what do you think about a release?

Please go ahead with #107, thanks! :smile:

I would like to address pytest-dev/pytest-mock#22 as well into `1.9.0`, if possible.

Issue #114, opened 2015-12-16
Hang when exception is raised in _quit_loop_by_signal


The-Compiler:
I haven't investigated that closely yet, thought I'd open an issue before I forget.

When raising an exception in `_quit_loop_by_signal`:

```diff
diff --git a/pytestqt/wait_signal.py b/pytestqt/wait_signal.py
index c32a76c..7efc019 100644
--- a/pytestqt/wait_signal.py
+++ b/pytestqt/wait_signal.py
@@ -111,6 +111,7 @@ class SignalBlocker(_AbstractSignalBlocker):
         """
         quits the event loop and marks that we finished because of a signal.
         """
+        raise Exception
         self.signal_triggered = True
         self._loop.quit()
         self._cleanup()
```

Then pytest-qt's tests seem to hang:

```
$ tox -e py35-pyqt5 -- tests/test_wait_signal.py -v 
[...]
test_wait_signal.py::test_signal_blocker_exception PASSED
test_wait_signal.py::test_signal_triggered[200-None-True-wait_function0-True] 
```

The-Compiler:
I think this is just happening because `self._loop.quit()` is never called, so we're stuck in the inner event loop, and handling the exception (which happened inside a virtual Qt function) will never be done.

I'm not sure there's anything good we can do about this (other than "don't raise an exception there" :laughing:), so if you can't think of anything either, I'm fine with closing this.

nicoddemus:
Perhaps we can make the code more robust to errors by wrapping it into a `try/finally`:

```python
         """
         quits the event loop and marks that we finished because of a signal.
         """
         try:
             self.signal_triggered = True
             self._cleanup()
         finally:
             self._loop.quit()
```

The-Compiler:
I thought about that too and thought it wasn't necessary because I can't imagine there was a condition where `self.signal_triggered = True` raises.

That being said, I think it's good to have it there as a "reminder" so any future additions (which could maybe raise) are in the `try` block, so :+1:.

nicoddemus:
> I think it's good to have it there as a "reminder" so any future additions

Yeah, that's exactly the intent, less for the current code and more as a reminder for future additions. Sorry for not being explicit. :smile:

Issue #113, opened 2015-11-29
Captured output leaking


The-Compiler:
I recently updated to pytest 2.8 and noticed I got output which should be captured on stdout.

This is during tests which launch a subprocess via `QProcess` which then does logging on stderr, which is printed during the tests (see [testprocess.py](https://github.com/The-Compiler/qutebrowser/blob/master/tests/integration/testprocess.py) / [quteprocess.py](https://github.com/The-Compiler/qutebrowser/blob/master/tests/integration/quteprocess.py#L156)).

I was able to reproduce this with this example:

```python
import pytest
from PyQt5.QtCore import QTimer


@pytest.yield_fixture(scope='module')
def timer():
    t = QTimer()
    t.setInterval(10)
    t.timeout.connect(lambda: print("foo"))
    t.start()
    yield t
    t.stop()


@pytest.mark.parametrize('x', range(1000))
def test_foo(qtbot, timer, x):
    qtbot.waitSignal(timer.timeout)
```

When running it with pytest 2.7.3:

```
============================================= test session starts ==============================================
platform linux -- Python 3.5.0 -- py-1.4.31 -- pytest-2.7.3
PyQt5 5.5.1 -- Qt runtime 5.5.1 -- Qt compiled 5.5.1
rootdir: /home/florian/proj/pytest, inifile: tox.ini
plugins: qt
collected 1000 items 

testcap2.py ...[…]...

========================================= 1000 passed in 1.16 seconds ==========================================
```

with 2.8.2:

```
============================================= test session starts ==============================================
platform linux -- Python 3.5.0, pytest-2.8.2, py-1.4.31, pluggy-0.3.1
PyQt5 5.5.1 -- Qt runtime 5.5.1 -- Qt compiled 5.5.1
rootdir: /home/florian/proj/pytest, inifile: tox.ini
plugins: qt-1.9.0
collected 1000 items 

testcap2.py .........foo
.........foo
.........foo
..........foo
...................foo
[…]

========================================= 1000 passed in 1.17 seconds ==========================================
```

I bisected this to https://github.com/pytest-dev/pytest/commit/c54afbe42e7e47afb7605f4bc226af92dbeda0a7 ("deprecate and warn about __multicall__ usage in hooks, refine docs about hook ordering, make hookwrappers respect tryfirst/trylast").

I also tried reproducing it without pytest-qt but it only seems to happen with pytest-qt.

[...]

PR #115, opened 2015-12-16
Add an 'args' attribute to SignalBlocker.

### LINKED ISSUE #64: Give access to what arguments were passed with in the signal
Thanks for the great plugin!

Would be awesome if when using qtbot.waitSignal, you could access any arguments that were sent in the signal e.g.

```python

with qtbot.waitSignal(worker.success) as blocker:
    pass
assert blocker.args == 'hello, world!'
```

What are your thoughts?

I'll take a stab at it if you think it's a good feature, thanks

The-Compiler:
Closes #64.

Maybe #114 needs a look first though - if there's anything wrong in the logic in `_quit_loop_by_signal` we'll just get a hang otherwise.

I have only tested on PyQt5.

### DIFF
--- a/docs/signals.rst
+++ b/docs/signals.rst
@@ -43,6 +43,22 @@ reached before the signal is triggered:
         assert_application_results(app)
 
 
+**Getting arguments of the emitted signal**
+
+.. versionadded:: 1.10
+
+The arguments emitted with the signal are available as the ``args`` attribute
+of the blocker:
+
+
+.. code-block:: python
+
+    def test_signal(qtbot):
+        ...
+        with qtbot.waitSignal(app.got_cmd) as blocker:
+            app.listen()
+        assert blocker.args == ['test']
+
 
 **waitSignals**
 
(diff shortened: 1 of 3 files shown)


The-Compiler:
> I have only tested on PyQt5

Seems like that means it only works on PyQt5 :laughing: 

Looks like PyQt4/PySide both don't have `QSignalSpy` available? I'll have to check how hard it'd be to do our own implementation of something similar instead.

The-Compiler:
Hah, looks like that's a lot easier and works everywhere.

nicoddemus:
Excellent, many thanks @The-Compiler! :smile:

I'm thinking of releasing `1.10` right away with this, what do you think?

The-Compiler:
After the adjustment from #114, sure, why not? I don't think I have anything I want to contribute in the next hours/days. I still need to look at #63 again. :laughing:

nicoddemus:
>  I don't think I have anything I want to contribute in the next hours/days.

Yeah, me too... pretty busy this week I'm afraid.

> I still need to look at #63 again. 

So do I. :disappointed:

The-Compiler:
No worries! The most annoying issue for me currently is #113, but I understand you're busy, and I have no idea how to tackle that one myself... :wink:

nicoddemus:
`1.10.0` is out. :smile:

The-Compiler:
Nice, thank you!

PR #119, opened 2015-12-17
Add qtbot.assertNotEmitted.

### LINKED ISSUE #92: assertSignal
I'd like to add a `qtbot.assertSignal` to pytest-qt

/cc @acogneau

# Motivation

Something I need to do regularily is to make sure a given code snippet calls a signal, or doesn't - but I don't want to wait for the signal.

The current way to do that works, but is unpythonic:

```python
spy = QSignalSpy(foo.signal)
foo.do_stuff()
assert len(spy) == 1
assert spy[1] == "signal_arg"
```

```python
spy = QSignalSpy(foo.error)
foo.do_stuff()
assert not spy
```

# API

After some thinking, I think a context manager with this signature would be best:

```python
qtbot.assertSignal(signal, count=None) -> QSignalSpy
```

* `signal`: The signal it should listen to
* `count`: Either `None` (in which case it expects the signal to be emitted >= 1 times), or an int. By setting `count` to `0`, it can be ensured the signal is *not* emitted.
* return value: A `QSignalSpy` which can be used to check the signal arguments.

# Examples

```python
with qtbot.assertSignal(foo.signal, count=1) as spy:
    foo.do_stuff()
assert spy[1] == "signal_arg"
```

```python
with qtbot.assertSignal(foo.error, count=0):
    foo.do_stuff()
```

# Alternatives

I came up with some other ideas while thinking about this, but I think all of them are worse:

* `qtbot.assertSignal(signal, emitted=True)`: Doesn't give you the possibility to check a signal was emitted exactly once, and isn't better in any way.
* `qtbot.assertSignal(signal, calls=[("signal arg 1", "signal arg 2")])`: Too complicated, and can be done easier by checking the `QSignalSpy`
* `qtbot.assertNotEmitted(signal)` - less functionality, and I think passing `count=0` is okay if you want that (that was the usecase I had in mind originally)
* Shoehorning this function into `waitSignal` - it's complex enough already, and `qtbot.waitSignal(signal, do_not_actually_wait=True)` just sounds wrong :wink:
* Adding a multi-signal variant (AND): Simply use `with qtbot.assertSignal(foo), qtbot.assertSignal(bar):` instead - that isn't an issue because it doesn't *wait* for the signal.
* Adding a multi-signal variant (OR): I don't really see the usecase for it.

The-Compiler:
I guess I lied when I said I wouldn't contribute anything smaller soon :laughing: 

Closes #92.

### DIFF
--- a/pytestqt/wait_signal.py
+++ b/pytestqt/wait_signal.py
@@ -188,7 +188,7 @@ class MultiSignalBlocker(_AbstractSignalBlocker):
 class SignalEmittedSpy(object):
 
     """
-    .. versionadded:: 1.4
+    .. versionadded:: 1.11
 
     An object which checks if a given signal has ever been emitted.
 
--- a/pytestqt/wait_signal.py
+++ b/pytestqt/wait_signal.py
@@ -185,12 +185,58 @@ class MultiSignalBlocker(_AbstractSignalBlocker):
         self._slots.clear()
 
 
+class SignalEmittedSpy(object):
+
+    """
+    .. versionadded:: 1.4
+
+    An object which checks if a given signal has ever been emitted.
+
+    Intended to be used as a context manager.
+    """
+
+    def __init__(self, signal):
+        self.signal = signal
+        self.emitted = False
+        self.args = None
+
+    def slot(self, *args):
+        self.emitted = True
+        self.args = args
+
+    def __enter__(self):
+        self.signal.connect(self.slot)
+
+    def __exit__(self, type, value, traceback):
+        self.signal.disconnect(self.slot)
+
+    def assert_not_emitted(self):
+        if self.emitted:
+            if self.args:
+                raise SignalEmittedError("Signal %r unexpectedly emitted with "
+                                         "arguments %r" %
+                                         (self.signal, list(self.args)))
+            else:
+                raise SignalEmittedError("Signal %r unexpectedly emitted" %
+                                         (self.signal,))
+
+
 class SignalTimeoutError(Exception):
     """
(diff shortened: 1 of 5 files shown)


tests/test_wait_signal.py
```diff
@@ -312,3 +313,34 @@ def test_connected_signal(self, qtbot, signaller):
...
+        with qtbot.assertNotEmitted(signaller.signal):
+            pass
+
+    def test_emitted(self, qtbot, signaller):
+        with pytest.raises(SignalEmittedError) as excinfo:
+            with qtbot.assertNotEmitted(signaller.signal):
+                signaller.signal.emit()
+
+        fnmatch.fnmatchcase(str(excinfo.value),
+                            "Signal * unexpectedly emitted.")
```
The-Compiler: Using `fnmatch` here because the `repr` of a signal is probably quite different between PyQt4/PyQt5/PySide.

pytestqt/wait_signal.py
```diff
@@ -185,6 +185,42 @@ def _cleanup(self):
         self._slots.clear()
 
 
+class SignalEmittedSpy(object):
+
+    """
+    .. versionadded:: 1.4
```
nicoddemus: 1.11?

The-Compiler: Err, of course :laughing:

nicoddemus:
Err sorry, what was holding this back again? Did I completely forgot about this PR by chance? :flushed:

The-Compiler:
Uh, I did as well :laughing: 

I don't remember anything being left to do here.

PR #121, opened 2016-01-06
[WIP] Add an option for the raising default

The-Compiler:
This is heavily work-in-progress (incomplete tests, `pytest.config` usage which is deprecated I think?), but already giving me headaches :wink:

It seems e.g. `TestArgs.test_timeout` fails, with `raising` set to `true` from the previous `testdir` run?!

```
___________________________________________________ TestArgs.test_timeout ____________________________________________________

self = <test_wait_signal.TestArgs object at 0x7f0c38c866d8>, qtbot = <pytestqt.qtbot.QtBot object at 0x7f0c38c865c0>

    def test_timeout(self, qtbot):
        """If there's a timeout, the args attribute is None."""
        with qtbot.waitSignal(timeout=100) as blocker:
>           pass

test_wait_signal.py:340: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../.tox/py35-pyqt5/lib/python3.5/site-packages/pytestqt/wait_signal.py:78: in __exit__
    self.wait()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pytestqt.wait_signal.SignalBlocker object at 0x7f0c3b779828>

    def wait(self):
        """
            Waits until either a connected signal is triggered or timeout is reached.
    
            :raise ValueError: if no signals are connected and timeout is None; in
                this case it would wait forever.
            """
        if self.signal_triggered:
            return
        if self.timeout is None and not self._signals:
            raise ValueError("No signals or timeout specified.")
        if self._timer is not None:
            self._timer.timeout.connect(self._quit_loop_by_timeout)
            self._timer.start()
        self._loop.exec_()
        if not self.signal_triggered and self.raising:
            print("raising: {!r}".format(self.raising))
            raise SignalTimeoutError("Didn't get signal after %sms." %
>                                    self.timeout)
E           pytestqt.wait_signal.SignalTimeoutError: Didn't get signal after 100ms.

[...]

PR #122, opened 2016-01-06
Add an option for the raising default

### LINKED ISSUE #120: Config to default to raising=True with waitSignal
I'd like a config option to default to `raising=True` for `waitSignal`/`waitSignals`, as I often forget adding it :wink:

nicoddemus:
The only trick was using `request.config.getini` instead of the deprecated `pytest.config` object... since it is a global instance and deprecated, it might have some corner case bugs, I don't know.

I was thinking about changing the default `raising` value to `True`. What do you think about warning users that `raising` will default to `True` in version `1.12`, and users should set `qt_wait_signal_raising` to `False` in `1.11` so when `1.12` is released their test suite doesn't break? Or perhaps `1.13`, if you feel one version span is too short?

Continuation from #121 
Fixes #120


nicoddemus:
Rebased :smile:

The-Compiler:
I think the way you implemented it (in `qtbot`), it won't work when using the blocker API by hand, right? That sounds quite surprising.

I think it would be good to have it as a default, yeah - the signal not being emitted being "ignored" by default certainly was a surprise to me. I don't have a strong opinion on a deprecation plan, I think your proposal sounds good. Maybe plan to do so with 1.12, and if we feel like the timespan between 1.11 and 1.12 was too short, push it back to 1.13?

nicoddemus:
> I think the way you implemented it (in qtbot), it won't work when using the blocker API by hand, right? That sounds quite surprising.

When I implemented this yesterday I thought that `SignalBlocker` and `MultiSignalBlocker` were not public API, but I just realized now that in fact they are. Two points:

1. I don't see how to implement the solution of passing the `request.config` object to `SignalBlocker` and `MultiSignalBlocker` without breaking its interface...
1. TBH I doubt they are much used; we could add a note to the documentation of the `raising` value saying that it ignores the value of the `qt_wait_signal_raising`. It is a little inconsistent, yes, but as I said I don't think it's much used and implementing it without using the problematic `pytest.config` is hard/impossible.

> Maybe plan to do so with 1.12, and if we feel like the timespan between 1.11 and 1.12 was too short, push it back to 1.13?

Sounds good. I will include that note into this PR then.

The-Compiler:
I wonder if I'm the only user of `SignalBlocker` :laughing: But yeah, I guess it's okay when it's documented accordingly.

nicoddemus:
> I wonder if I'm the only user of SignalBlocker

Oh do you? :grimacing:

Well, I will see if it's possible then... do you have any suggestions, perhaps?

The-Compiler:
Yeah, in one place where I need to wait for a signal in a module-scoped fixture.

Unfortunately I don't have a good idea. As you said, any implementation not involving `pytest.config` would mean the caller would have to pass the config, which is kind of pointless.

nicoddemus:
Yeah, me neither.

I will add a note to the docs of `SignalBlocker` and `MultiSignalBlocker` about they not using the config option, add a note about changing the default value in a future version. How does that sound?

The-Compiler:
:+1:

PR #123, opened 2016-02-07
Fix syntax of `tests_require` option

estan:
Confusingly, the option has no trailing 's' like `install_requires`.

See:

* https://pythonhosted.org/setuptools/setuptools.html
* https://travis-ci.org/pytest-dev/pytest-qt/jobs/100969573#L363


The-Compiler:
Thank you!

PR #124, opened 2016-02-16
Use pytest-xvfb.

The-Compiler:
This prevents windows popping up when running the tests on Linux.

Also, shameless promotion of [the plugin I just wrote](https://github.com/The-Compiler/pytest-xvfb), annoucement mail will follow soon :wink:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
 
 script:
  - python setup.py develop
- - catchsegv xvfb-run coverage run --source=pytestqt setup.py test
+ - catchsegv coverage run --source=pytestqt setup.py test
 
 after_success:
  - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
  - pip install -U pytest
 
  # others
- - pip install coveralls --use-wheel
+ - pip install coveralls pytest-xvfb --use-wheel
 
 script:
  - python setup.py develop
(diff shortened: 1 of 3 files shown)


nicoddemus:
Thanks! Would you like to contribute a section to the docs to suggest using your plugin as well? I think it is completely relevant!

The-Compiler:
Sure - didn't think of that!

Where do you think it'd fit in? The best place I found so far is in `intro.rst` in a new "Related plugins" section or so. Any better idea?

nicoddemus:
Sounds good, go for it.

The-Compiler:
Done!

The-Compiler:
Weird, some Travis environments are failing after the tests are finished:

```
=================== 123 passed, 2 skipped in 17.15 seconds ====================
plugin.py: Fatal IO error: client killed
```

Right now I have no idea where that message is coming from...

nicoddemus:
No idea either... I assume you are using it in `qutebrowser` without any problems?

nicoddemus:
A [quick](https://github.com/thoughtbot/capybara-webkit/issues/356#issuecomment-11791430) (really quick as I have to go to bed) suggests `xvfb` is being killed before tests are finished with it? Perhaps the fact that `qApp` never really "shuts down"?

The-Compiler:
Yeah, I came to the same conclusion: https://github.com/The-Compiler/pytest-xvfb/issues/1

No issues with qutebrowser, I'm guessing it works with PyQt5 or something... No idea what I could do to fix this yet :unamused:

nicoddemus:
:disappointed:

nicoddemus:
Let's keep this open then until we figure out a solution

The-Compiler:
I guess there's no point in keeping this open before someone has some kind of epiphany for https://github.com/The-Compiler/pytest-xvfb/issues/1 😆

PR #126, opened 2016-04-05
When an exception is caught in sys.excepthook, still print it to stderr.

fabioz:


tests/test_exceptions.py
```diff
@@ -244,3 +244,35 @@ def event(self, ev):
         qapp.processEvents()
 
     assert [str(e) for (t, e, tb) in exceptions] == ['mistakes were made']
+
+
+def test_exceptions_to_stderr(qapp):
```
nicoddemus: Please use the [capsys](http://pytest.org/latest/capture.html?highlight=capsys#accessing-captured-output-from-a-test-function) fixture instead, so you will avoid having to manually override `sys.stderr`.

fabioz:
Using capsys seems to have fixed the errors... the current errors (both in appveyor and in coveralls) don't seem related to my change.

nicoddemus:
Thanks, it seems it is because of something on the tox env which checks the docs. I will take a look later.

PR #130, opened 2016-05-12
Configure Qt API using config file

### LINKED ISSUE #129: Configuring Qt API in pytest.ini
In `#pylib` I discussed with @estan if there's some good way to configure `PYTEST_QT_API` in a way which doesn't require passing it on the commandline, for cases where you always need it to be e.g. `pyqt5`.

Unfortunately using [pytest-env](https://github.com/MobileDynasty/pytest-env) doesn't work as it uses `pytest_load_initial_conftestst` which will be run after `pytest-qt` is already imported.

One possibility would be to change `pytest-qt` so it reads the API version from `pytest.ini`. This should be doable but is probably quite painful, as it reqires delaying any imports from `qt_compat` until after `pytest_load_initial_conftests` (i.e. moving a lot of imports inside functions...).

Another possibility would be to add an `env` option to pytest's core, but pytest would need to read it (and set the environment variables) before any plugins are loaded...

Alternatively `pytest-qt` could search a `pytestqt.ini` or so in the current directory, but that'd also break in various circumstances and isn't really nice either.

What do you think?

nicoddemus:
Fix #129

The big work here was of course to implement lazy-loading of Qt classes from `qt_compat`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,8 +37,19 @@ New Features
   is met or a timeout is reached. Useful for testing some features in X window environments
   due to its asynchronous nature.
 
+* Now which Qt binding ``pytest-qt`` should use can be configured by the ``qt_api`` config option.
+  Thanks `@The-Compiler`_ for the request (`#129`_).
+
+- ``PYTEST_QT_FORCE_PYQT`` environment variable is no longer supported.
+
+- While ``pytestqt.qt_compat`` is an internal module and shouldn't be imported directly,
+  it is known that some test suites did import it. This module now uses a lazy-load mechanism
+  to load Qt classes and objects, so the old symbols (``QtCore``, ``QApplication``, etc.) are
+  no longer available from it.
+
 .. _#134: https://github.com/pytest-dev/pytest-qt/issues/134
 .. _#63: https://github.com/pytest-dev/pytest-qt/pull/63
+.. _#129: https://github.com/pytest-dev/pytest-qt/issues/129
 
 
 Other Changes
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,9 @@ However, this required some backwards-incompatible changes:
   on timeouts. You can set ``qt_wait_signal_raising = false`` in your config to
   get back the old behaviour.
 
+- ``PYTEST_QT_FORCE_PYQT`` environment variable is no longer supported. Set ``PYTEST_QT_API``
+  to the appropriate value instead.
+
 
 New Features
 ~~~~~~~~~~~~
@@ -39,8 +42,6 @@ New Features
 
 * Now which Qt binding ``pytest-qt`` should use can be configured by the ``qt_api`` config option.
   Thanks `@The-Compiler`_ for the request (`#129`_).
-
-- ``PYTEST_QT_FORCE_PYQT`` environment variable is no longer supported.
 
 - While ``pytestqt.qt_compat`` is an internal module and shouldn't be imported directly,
   it is known that some test suites did import it. This module now uses a lazy-load mechanism
--- a/pytestqt/modeltest.py
+++ b/pytestqt/modeltest.py
@@ -34,8 +34,7 @@
 from __future__ import print_function
 import collections
 
-from pytestqt.qt_compat import QtCore, QtGui, extract_from_variant, \
-    QAbstractListModel, QAbstractTableModel
+from pytestqt.qt_compat import qt_api
 
 
 _Changing = collections.namedtuple('_Changing', 'parent, old_size, last, next')
@@ -65,10 +64,10 @@ class ModelTester:
         if not index.isValid():
             return '<invalid> (0x{:x})'.format(id(index))
         else:
-            data = self._model.data(index, QtCore.Qt.DisplayRole)
+            data = self._model.data(index, qt_api.QtCore.Qt.DisplayRole)
             return '{}/{} {!r} (0x{:x})'.format(
                 index.row(), index.column(),
-                extract_from_variant(data),
+                qt_api.extract_from_variant(data),
                 id(index))
 
     def check(self, model):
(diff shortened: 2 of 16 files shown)


The-Compiler:
Oh wow, crazy stuff! :laughing: :+1:

What do you think about setting a flag when `pytest_configure` was executed (or maybe pytest has already one?), and then checking that when doing the actual Qt import (to make sure no import happened before the config was read)?

As for testing, I guess you could set up an environment with multiple Qt libraries installed so pytest-qt would pick the wrong one, then run in a subprocess (to make sure nothing is imported) with the config option, and e.g. look at the version output (or simply run a test in the subprocess) to make sure the right Qt API was picked?

nicoddemus:
> What do you think about setting a flag when pytest_configure was executed (or maybe pytest has already one?), and then checking that when doing the actual Qt import (to make sure no import happened before the config was read)?

That does not seem necessary: all imports will happen in the `set_qt_api` method, and that method is called in `pytest_configure` already. Or do you see some scenario or that might not be True?

> As for testing, I guess you could set up an environment with multiple Qt libraries installed so pytest-qt would pick the wrong one, then run in a subprocess (to make sure nothing is imported) with the config option, and e.g. look at the version output (or simply run a test in the subprocess) to make sure the right Qt API was picked?

Good idea, although I it will complicate the tox setup. But you gave me an idea: setting the ini variable to the *wrong* api should raise an appropriate `ImportError` when the plugin loads. So, in a `PySide` environment this ini file:

```ini
[pytest]
qt_api=pyqt4
```

Will fail make pytest fail with an `ImportError: PyQt4`, thus proving the variable is working. What do you think?

## qt_compat

This change will break backward compatibility for the `qt_compat` module, since `from pytestqt.qt_compat import QApplication` (for example) will not work anymore. I think this will have to warrant a `2.0` release to follow semantic versioning consistently, even thought I don't think it will break much code out there. I will probably rename it to `_qt_compat` to remove it from the public API as well. Thoughts?

The-Compiler:
> > What do you think about setting a flag when pytest_configure was executed (or maybe pytest has already one?), and then checking that when doing the actual Qt import (to make sure no import happened before the config was read)?
> 
> That does not seem necessary: all imports will happen in the `set_qt_api` method, and that method is called in `pytest_configure` already. Or do you see some scenario or that might not be True?

You're right! I thought using `qt_api.QtFoo` would automatically import it no matter what, but I didn't take a close enough look.

Though a test which makes sure importing before `pytest_configure` fails would be nice.

> > As for testing, I guess you could set up an environment with multiple Qt libraries installed so pytest-qt would pick the wrong one, then run in a subprocess (to make sure nothing is imported) with the config option, and e.g. look at the version output (or simply run a test in the subprocess) to make sure the right Qt API was picked?
> 
> Good idea, although I it will complicate the tox setup. But you gave me an idea: setting the ini variable to the *wrong* api should raise an appropriate `ImportError` when the plugin loads. So, in a `PySide` environment this ini file:
> 
> ```ini
> [pytest]
> qt_api=pyqt4
> ```
> 
> Will fail make pytest fail with an `ImportError: PyQt4`, thus proving the variable is working. What do you think?

Sounds good to me.

> ## qt_compat
> 
> This change will break backward compatibility for the `qt_compat` module, since `from pytestqt.qt_compat import QApplication` (for example) will not work anymore. I think this will have to warrant a `2.0` release to follow semantic versioning consistently, even thought I don't think it will break much code out there. I will probably rename it to `_qt_compat` to remove it from the public API as well. Thoughts?

Agreed. Is there anything else which would go into a 2.0 release? What about delaying 2.0 until after the pytest sprint? Then we can discuss if there's anything else we'd like to change which would need an API break (for example, I'd like to discuss making `qt_wait_signal_raising` default to `True`).

I also thought about if we should use/contribute to one of the existing wrappers instead, like [`pyqode.qt`](https://github.com/pyQode/pyqode.qt) - but probably it's not worth the trouble, as I'd guess @ColinDuquesnoy is more interested in the needs of [pyQode](https://github.com/pyQode/pyQode) (for which it was written for) :wink:

nicoddemus:
> Agreed. Is there anything else which would go into a 2.0 release? What about delaying 2.0 until after the pytest sprint? 

Sure, sounds good!

> for example, I'd like to discuss making qt_wait_signal_raising default to True

No discussion there! :wink: In fact it was supposed to change even before that, as stated [in this note](https://pytest-qt.readthedocs.io/en/latest/changelog.html#id3). But I agree that changing it into a 2.0 is better specially if it is not far away.

> I also thought about if we should use/contribute to one of the existing wrappers instead, like pyqode.qt - but probably it's not worth the trouble

I also don't think it would be worth it: pytest-qt uses a very limited set of the Qt api that just our humble `qt_compat` module works well, I think (although a little ugly), and I wouldn't like to add an extra dependency without a very compelling reason.

ColinDuquesnoy:
> I also thought about if we should use/contribute to one of the existing wrappers instead, like pyqode.qt - but probably it's not worth the trouble, as I'd guess @ColinDuquesnoy is more interested in the needs of pyQode (for which it was written for) :wink:

I also think it wouldn't be worth it. There are already many wrappers, nearly any project that wants to support pyside, pyqt4 and pyqt5 have their own wrappers, written for their own needs.

tests/test_basics.py
```diff
@@ -270,9 +269,42 @@ def event_recorder(qtbot):
...
 
 
 def test_parse_ini_boolean_invalid():
+    import pytestqt.qtbot
     with pytest.raises(ValueError):
         pytestqt.qtbot._parse_ini_boolean('foo')
+
+
+@pytest.mark.parametrize('option_api', ['pyqt4', 'pyqt4v2', 'pyqt5', 'pyside'])
+def test_qt_api_ini_config(testdir, option_api):
```
nicoddemus: This is how I ended up testing the `qt_api` variable in the end

nicoddemus:
@The-Compiler rebased and I think this is ready for review. Could you take a look when you have the time please? :grin:

CHANGELOG.rst
```diff
@@ -37,8 +37,19 @@ New Features
   is met or a timeout is reached. Useful for testing some features in X window environments
   due to its asynchronous nature.
 
+* Now which Qt binding ``pytest-qt`` should use can be configured by the ``qt_api`` config option.
+  Thanks `@The-Compiler`_ for the request (`#129`_).
+
+- ``PYTEST_QT_FORCE_PYQT`` environment variable is no longer supported.
```
The-Compiler: Shouldn't this be under "breaking changes"?

pytestqt/modeltest.py
```diff
@@ -636,44 +635,44 @@ def _on_data_changed(self, top_left, bottom_right):
...
             item_count = self._column_count()
         assert start < item_count
         assert end < item_count
 
-    def _column_count(self, parent=QtCore.QModelIndex()):
+    def _column_count(self, parent=qt_api.QtCore.QModelIndex()):
         """
         Workaround for the fact that ``columnCount`` is a private method in
-        QAbstractListModel/QAbstractTableModel subclasses.
+        qt_api.QAbstractListModel/qt_api.QAbstractTableModel subclasses.
```
The-Compiler: I don't think the `qt_api.` prefix is needed here.

pytestqt/qt_compat.py
```diff
@@ -13,211 +13,181 @@
...
-            return True
-        except ImportError:
-            return False
+class _QtApi:
+    """
+    Interface to the underlying Qt API currently configured for pytest-qt.
 
-    def _guess_qt_api():
-        if _try_import('PyQt5'):
+    This object lazily loads all class references and other objects upon the ``set_qt_api`` method,
```
The-Compiler: "upon the ``set_qt_api`` method" sounds a bit strange - maybe "when the ``set_qt_api`` method is called"?

CHANGELOG.rst
nicoddemus: Oops, bad merge, thanks!

pytestqt/qt_compat.py
```diff
@@ -13,211 +13,181 @@
...
+        _root_modules = {
+            'pyside': 'PySide',
+            'pyqt4': 'PyQt4',
+            'pyqt4v2': 'PyQt4',
+            'pyqt5': 'PyQt5',
+        }
+        _root_module = _root_modules[self.pytest_qt_api]
+
+        def _import_module(module_name):
+            m = __import__(_root_module, globals(), locals(), [module_name], 0)
```
The-Compiler: crazy stuff :laughing:

nicoddemus: what can I say, lazily loading stuff usually is :stuck_out_tongue_closed_eyes:

tests/test_qtest_proxies.py
```diff
@@ -1,10 +1,10 @@
 import pytest
 
-from pytestqt.qt_compat import USING_PYSIDE
+# noinspection PyUnresolvedReferences
```
The-Compiler: Just curious - is this a PyCharm thing?

nicoddemus: Yes, otherwise PyCharm complains this is not being used. It's was complaining because the mark is now "eval"ued instead of being explicit.

The-Compiler:
LGTM otherwise, let's see what AppVeyor/Travis think :wink:

The-Compiler:
One thing I'm still not sure about: What if someone does `from pytestqt.qt_api import QtCore`? Will they get an `AttributeError`? Maybe it'd be worth adding a test for that?

nicoddemus:
`from pytestqt.qt_api import QtCore` will get you an `ImportError` because there isn't a `qt_api` module. :stuck_out_tongue_closed_eyes: 

But I see what you mean. When someone does a `from pytestqt.qt_compat.qt_api import QtCore` they will get an `ImportError` I guess. Don't see much the point of adding a test for that, TBH

The-Compiler:
Fair enough :wink:

This happens: `ImportError: No module named 'pytestqt.qt_compat.qt_api'; 'pytestqt.qt_compat' is not a package`

Seems like one line [is not covered anymore](https://coveralls.io/builds/6814711/source?filename=pytestqt%2Fqt_compat.py#L143), or did coveralls just not update yet?

nicoddemus:
>  or did coveralls just not update yet?

Not covered yet, it seems OK since the [last coverage report](https://coveralls.io/builds/6814711).

But now I'm not sure why [this](https://coveralls.io/builds/6814711/source?filename=pytestqt%2Fqt_compat.py#L143) is not getting covered, after all we have a travis build specifically for `pyqt4v2`...

The-Compiler:
There's a `if self.pytest_qt_api == 'pyqt5':` [above](https://github.com/pytest-dev/pytest-qt/pull/130/files#diff-9e4e17fe090e82f219f1f4d91e0e70d8R134) if we're talking about the same thing :wink:

pytestqt/qt_compat.py
```diff
@@ -13,211 +13,181 @@
...
+            return getattr(m, module_name)
+
+        if self.pytest_qt_api == 'pyqt4v2':
```
nicoddemus: @The-Compiler actually I mean this one...

The-Compiler:
heh, I guess that works too :laughing:

nicoddemus:
I really don't have a clue why those lines are reported as missing... :sob:

The-Compiler:
Are they covered when you do something like `PYTEST_QT_API=pyqt4v2 coverage run --source=pytestqt py.test ; coverage html` locally?

I'll take another look tomorrow.

nicoddemus:
@The-Compiler, I plan to merge this on the weekend after I figure out why coverage decreased. 

Just wanted to know if you will find the time to get `waitCallback` (is that what we decided to call it?) into `2.0` as well? Otherwise after this PR is merged we can make the release. :grin:

The-Compiler:
Probably not, I'd expect to work on it in a few days to weeks - so let's release 2.0 without it and then do a 2.1 at some point :wink:

nicoddemus:
Roger! :+1:

nicoddemus:
Duh, I was looking at the wrong line.

Will merge after #141 to apply the proper changes to that branch myself.

estan:
@nicoddemus @The-Compiler: Thanks! The new option works great. I was the one who initially bugged @The-Compiler on IRC about this :)

nicoddemus:
:+1:

PR #139, opened 2016-06-29
Add waitUntil method to QtBot

nicoddemus:
As we discussed during the sprint, adding a new `qtbot.waitUntil` method.

### DIFF
--- a/pytestqt/qtbot.py
+++ b/pytestqt/qtbot.py
@@ -1,7 +1,8 @@
 import functools
 import contextlib
 import weakref
-from pytestqt.wait_signal import SignalBlocker, MultiSignalBlocker, SignalTimeoutError, SignalEmittedSpy
+from pytestqt.wait_signal import SignalBlocker, MultiSignalBlocker, SignalTimeoutError, \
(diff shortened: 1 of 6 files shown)


pytestqt/qtbot.py
```diff
@@ -373,6 +375,73 @@ def assertNotEmitted(self, signal):
...
+                # 'assert' form
+                if result is None:
+                    return
+
+                # 'True/False' form
+                if result:
+                    return
+                else:
+                    assert not timed_out(), 'waitUntil timed out in %s miliseconds' % timeout
+            self.wait(10)
```
The-Compiler: Maybe this should be an argument?

nicoddemus: Thought about it, but didn't think it would end up being necessary... perhaps we should wait until someone asks for it?

The-Compiler: Sounds like a plan :wink:

nicoddemus:
Merge? :grin:

The-Compiler:
Yep! :+1:

Issue #118, opened 2015-12-17
Waiting for a signal with certain arguments


The-Compiler:
I have some scenarios where I'd like to do something like:

```python
with qtbot.waitSignal(got_line, args=['foo']):
    ...
```

And qtbot would wait until the signal is emitted with exactly those args.

I also have another case where I do some more sophisticated checking:

```python
    spy = QSignalSpy(self.new_data)
    elapsed_timer = QElapsedTimer()
    elapsed_timer.start()

    while True:
        got_signal = spy.wait(timeout)
        if not got_signal or elapsed_timer.hasExpired(timeout):
            raise WaitForTimeout("Timed out after {}ms waiting for "
                                 "{!r}.".format(timeout, kwargs))

        for args in spy:
            assert len(args) == 1
            line = args[0]

            matches = []

            for key, expected in kwargs.items():
                value = getattr(line, key)
                matches.append(self._match_data(value, expected))

            if all(matches):
                return line
```

I'm thinking if it makes sense or not to do something like:

```python
def check_args(*args):
    return args == ['foo', 2342]

with qtbot.waitSignal(got_line, callback=check_args):
    ...
```

Then `qtbot` will only stop waiting when the callback returned something true.

Maybe `args` isn't even necessary then, as you could do:

```python
with qtbot.waitSignal(got_line, callback=lambda *args: args == ['foo']):
    ...
```

What do you think? Neither? Only `args`? `args` and `callback`? Both?

(edited to rename `arg_callback` to `callback` as it could be used to check other stuff as well.

nicoddemus:
I would add only `callback` as it is more generic and would cover most use cases...

MShekow:
Hi. I've had a shot at it, because I needed the ability to check for parameters (of a single signal or even multiple signals), and I also needed the ability to check for a correct order of emitted signals, and also the ability to check that the same signal is emitted in a specific order (each emit with a different parameter).

Here are my method signatures:
Single signal:
`waitSignal(self, signal=None, timeout=1000, callback=None, raising=None):`
As suggested above. The docblock says:
`:param Callable callback:
            Optional callable(*args) that returns True if the parameters are as expected, or False otherwise.`

For multiple signals:
`def waitSignals(self, signals=None, timeout=1000, force_order = "none", callbacks = None, raising=None):`
[...]

PR #141, opened 2016-07-14
waitSignal and waitSignals() evaluate signal parameters

MShekow:
waitSignal and waitSignals() have a callback/callbacks parameter that makes it possible to test that emitted signals have specific parameters set. waitSignals() also has a force_order parameter that allows to configure 3 different levels of strictness regarding the order in which the signals are expected. See https://github.com/pytest-dev/pytest-qt/issues/118

### DIFF
--- a/pytestqt/qtbot.py
+++ b/pytestqt/qtbot.py
@@ -234,7 +234,7 @@ class QtBot(object):
 
     stop = stopForInteraction
 
-    def waitSignal(self, signal=None, timeout=1000, raising=None):
+    def waitSignal(self, signal=None, timeout=1000, callback=None, raising=None):
         """
         .. versionadded:: 1.2
 
@@ -265,6 +265,8 @@ class QtBot(object):
             A signal to wait for. Set to ``None`` to just use timeout.
         :param int timeout:
             How many milliseconds to wait before resuming control flow.
+        :param Callable callback:
+            Optional callable(*args) that returns True if the parameters are as expected, or False otherwise.
         :param bool raising:
             If :class:`QtBot.SignalTimeoutError <pytestqt.plugin.SignalTimeoutError>`
             should be raised if a timeout occurred.
@@ -285,14 +287,14 @@ class QtBot(object):
                 raising = True
             else:
                 raising = _parse_ini_boolean(raising_val)
-        blocker = SignalBlocker(timeout=timeout, raising=raising)
+        blocker = SignalBlocker(timeout=timeout, raising=raising, callback=callback)
         if signal is not None:
             blocker.connect(signal)
         return blocker
 
     wait_signal = waitSignal  # pep-8 alias
 
-    def waitSignals(self, signals=None, timeout=1000, raising=None):
+    def waitSignals(self, signals=None, timeout=1000, force_order = "none", callbacks = None, raising=None):
         """
         .. versionadded:: 1.4
 
@@ -315,10 +317,19 @@ class QtBot(object):
            blocker.wait()
 
         :param list signals:
-            A list of :class:`Signal` objects to wait for. Set to ``None`` to
-            just use timeout.
+            A list of :class:`Signal`s objects to wait for. Set to ``None`` to just use
+            timeout.
         :param int timeout:
             How many milliseconds to wait before resuming control flow.
+        :param str force_order:
+            "none": no order is enforced
+            "strict": the signals have to be emitted in the provided order (and no intermediate signals may be emitted)
+            "simple": like "strict", but signals may be emitted inbetween the provided ones, e.g. suppose
+            signals=[a, b, c] and the tested object emits the signals a, a, b, a, c, the test will pass with "simple"
+            but not with "strict"
+        :param list callbacks:
+            optional list of Callable functions that evaluate the parameters of the signal. Each element of the list
+            can be a callable or None. Make sure that the number of callbacks matches the number of signals.
         :param bool raising:
             If :class:`QtBot.SignalTimeoutError <pytestqt.plugin.SignalTimeoutError>`
             should be raised if a timeout occurred.
@@ -334,12 +345,18 @@ class QtBot(object):
 
         .. note:: This method is also available as ``wait_signals`` (pep-8 alias)
         """
+        assert force_order in ["none", "simple", "strict"], "force_order has to be set to 'none', 'simple' or 'strict'"
+
         if raising is None:
             raising = self._request.config.getini('qt_wait_signal_raising')
-        blocker = MultiSignalBlocker(timeout=timeout, raising=raising)
+
+        if callbacks:
+            if len(callbacks) != len(signals):
+                raise ValueError("Number of callbacks ({}) does not "
+                                 "match number of signals ({})!".format(len(callbacks), len(signals)))
+        blocker = MultiSignalBlocker(timeout=timeout, raising=raising, force_order=force_order, callbacks=callbacks)
         if signals is not None:
-            for signal in signals:
-                blocker._add_signal(signal)
+            blocker.add_signals(signals)
         return blocker
 
     wait_signals = waitSignals  # pep-8 alias
--- a/pytestqt/qtbot.py
+++ b/pytestqt/qtbot.py
@@ -234,7 +234,7 @@ class QtBot(object):
 
     stop = stopForInteraction
 
-    def waitSignal(self, signal=None, timeout=1000, callback=None, raising=None):
+    def waitSignal(self, signal=None, timeout=1000, raising=None, check_params_cb=None):
         """
         .. versionadded:: 1.2
 
@@ -265,13 +265,15 @@ class QtBot(object):
             A signal to wait for. Set to ``None`` to just use timeout.
         :param int timeout:
             How many milliseconds to wait before resuming control flow.
-        :param Callable callback:
-            Optional callable(*args) that returns True if the parameters are as expected, or False otherwise.
         :param bool raising:
             If :class:`QtBot.SignalTimeoutError <pytestqt.plugin.SignalTimeoutError>`
             should be raised if a timeout occurred.
             This defaults to ``True`` unless ``qt_wait_signal_raising = false``
             is set in the config.
+        :param Callable check_params_cb:
+            Optional callable(*parameters) that compares the provided signal parameters to some expected parameters.
+            It has to match the signature of ``signal`` (just like a slot function would) and return ``True`` if
+            parameters match, ``False`` otherwise.
         :returns:
             ``SignalBlocker`` object. Call ``SignalBlocker.wait()`` to wait.
 
@@ -287,14 +289,14 @@ class QtBot(object):
                 raising = True
             else:
                 raising = _parse_ini_boolean(raising_val)
-        blocker = SignalBlocker(timeout=timeout, raising=raising, callback=callback)
+        blocker = SignalBlocker(timeout=timeout, raising=raising, check_params_cb=check_params_cb)
         if signal is not None:
             blocker.connect(signal)
         return blocker
 
     wait_signal = waitSignal  # pep-8 alias
 
-    def waitSignals(self, signals=None, timeout=1000, force_order = "none", callbacks = None, raising=None):
+    def waitSignals(self, signals=None, timeout=1000, raising=None, check_params_cbs=None, order="none"):
         """
         .. versionadded:: 1.4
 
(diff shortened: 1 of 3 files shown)


pytestqt/qtbot.py
```diff
@@ -234,7 +234,7 @@ def stopForInteraction(self):
 
     stop = stopForInteraction
 
-    def waitSignal(self, signal=None, timeout=1000, raising=None):
+    def waitSignal(self, signal=None, timeout=1000, callback=None, raising=None):
```
nicoddemus: I think this parameter should appear at the end, so code doing `waitSignal(signal, 1000, True)` won't break.

nicoddemus: Also I feel `callback` is to generic... perhaps `check_params_callback` would convey the intent a little better?

nicoddemus:
Hey @MShekow thanks a lot for the PR.

I have to leave to work so I can't review it more carefully now, I will try to make a proper review during lunch, or at the end the day the latest.

pytestqt/qtbot.py
MShekow: I don't mind this rename.

The-Compiler:
hmm, no idea why the windows builds are broken...

pytestqt/qtbot.py
```diff
@@ -285,14 +287,14 @@ def waitSignal(self, signal=None, timeout=1000, raising=None):
...
-        blocker = SignalBlocker(timeout=timeout, raising=raising)
+        blocker = SignalBlocker(timeout=timeout, raising=raising, callback=callback)
         if signal is not None:
             blocker.connect(signal)
         return blocker
 
     wait_signal = waitSignal  # pep-8 alias
 
-    def waitSignals(self, signals=None, timeout=1000, raising=None):
+    def waitSignals(self, signals=None, timeout=1000, force_order = "none", callbacks = None, raising=None):
```
The-Compiler: nitpick: No spaces around default values

pytestqt/qtbot.py
```diff
@@ -315,10 +317,19 @@ def waitSignals(self, signals=None, timeout=1000, raising=None):
            blocker.wait()
 
         :param list signals:
-            A list of :class:`Signal` objects to wait for. Set to ``None`` to
-            just use timeout.
+            A list of :class:`Signal`s objects to wait for. Set to ``None`` to just use
```
The-Compiler: That added "s" is wrong :wink:

pytestqt/qtbot.py
The-Compiler: Also I think I'd rename `force_order` to `order`, and the parameters should be at the end as well.

pytestqt/qtbot.py
The-Compiler: `check_params_cb` maybe? I think `_cb` is a common suffix, and `check_params_callback` seems so long :wink:

pytestqt/qtbot.py
```diff
@@ -315,10 +317,19 @@ def waitSignals(self, signals=None, timeout=1000, raising=None):
...
         :param int timeout:
             How many milliseconds to wait before resuming control flow.
+        :param str force_order:
+            "none": no order is enforced
+            "strict": the signals have to be emitted in the provided order (and no intermediate signals may be emitted)
+            "simple": like "strict", but signals may be emitted inbetween the provided ones, e.g. suppose
+            signals=[a, b, c] and the tested object emits the signals a, a, b, a, c, the test will pass with "simple"
+            but not with "strict"
+        :param list callbacks:
+            optional list of Callable functions that evaluate the parameters of the signal. Each element of the list
```
The-Compiler: "Callable functions" sounds weird - either "list of callables" or "list of functions" IMHO.

pytestqt/qtbot.py
```diff
@@ -334,12 +345,18 @@ def waitSignals(self, signals=None, timeout=1000, raising=None):
 
         .. note:: This method is also available as ``wait_signals`` (pep-8 alias)
         """
+        assert force_order in ["none", "simple", "strict"], "force_order has to be set to 'none', 'simple' or 'strict'"
```
The-Compiler: This should be a `if` raising a `ValueError` - assertions are deactivated with `-O` for example, and an `AssertionError` is a weird thing to get for invalid parameters.

pytestqt/wait_signal.py
```diff
@@ -151,48 +152,121 @@ class MultiSignalBlocker(_AbstractSignalBlocker):
...
             _silent_disconnect(signal, slot)
-        self._signals.clear()
-        self._slots.clear()
+        del self._signals_emitted[:]
+        self._signals_map.clear()
+        del self._slots[:]
 
 
 class SignalEmittedSpy(object):
-
```
The-Compiler: Unrelated whitespace change

tests/test_wait_signal.py
```diff
@@ -335,7 +336,6 @@ class Obj(QtCore.QObject):
 
 
 class TestArgs:
-
```
The-Compiler: Lots of unrelated whitespace changes :wink:

MShekow: These line changes must have happened when I was using PyCharm's **Reformat code** functionality, which cleans up files very nicely. I'm not familiar with the general rule for this project. Normally, a project should define an open source tool that can be executed on a python module and it will format it according to rules set up by the project team. Is such a mechanism in place?

pytestqt/qtbot.py
MShekow: I like "list of callables", as "callable" is a Python term.

pytestqt/qtbot.py
MShekow: I'm OK with renaming the parameter to `order` and placing it at the end. Spaces around the default value of the parameter can be removed.

MShekow:
As I've never done PRs, I'd like to know whether the general procedure is this one?:
You suggest changes to the code, I add them to my fork, and once you're satisfied, you merge the PR?

tests/test_wait_signal.py
The-Compiler: I don't really know what @nicoddemus uses - but either way I personally think doing cleanup changes as part of other PRs is problematic (when using git blame, when having to revert something, etc.). I think @nicoddemus is a bit more relaxed than I in that regard though :laughing:

The-Compiler:
Right - if you push new changes to your fork, the PR will update automatically. Then we review it again, you update it if needed, etc., until it gets merged.

The-Compiler:
FWIW I get the same AppVeyor issue [for qutebrowser as well](https://ci.appveyor.com/project/The-Compiler/qutebrowser/build/master-3152/job/21wndp808hofslqx). I'll investigate tomorrow.

pytestqt/qtbot.py
```diff
@@ -315,10 +317,19 @@ def waitSignals(self, signals=None, timeout=1000, raising=None):
...
 
         :param list signals:
-            A list of :class:`Signal` objects to wait for. Set to ``None`` to
-            just use timeout.
+            A list of :class:`Signal`s objects to wait for. Set to ``None`` to just use
+            timeout.
         :param int timeout:
             How many milliseconds to wait before resuming control flow.
+        :param str force_order:
+            "none": no order is enforced
```
nicoddemus: Please format those as a list, like in [here](https://github.com/pytest-dev/pytest-qt/blob/master/pytestqt/qtbot.py#L109). The results are [worth it](http://pytest-qt.readthedocs.io/en/1.11.0/reference.html#pytestqt.qtbot.QtBot.mouseRelease) IMO.

pytestqt/qtbot.py
nicoddemus: It is not very clear from the docs here what the callbacks are supposed to do with the signals they receive.

pytestqt/qtbot.py
```diff
@@ -315,10 +317,19 @@ def waitSignals(self, signals=None, timeout=1000, raising=None):
...
+            "strict": the signals have to be emitted in the provided order (and no intermediate signals may be emitted)
+            "simple": like "strict", but signals may be emitted inbetween the provided ones, e.g. suppose
+            signals=[a, b, c] and the tested object emits the signals a, a, b, a, c, the test will pass with "simple"
+            but not with "strict"
+        :param list callbacks:
+            optional list of Callable functions that evaluate the parameters of the signal. Each element of the list
+            can be a callable or None. Make sure that the number of callbacks matches the number of signals.
```
nicoddemus: > Make sure that the number of callbacks matches the number of signals.

I suggest make it explicit about the `ValueError`, something like: 

``` 
If the number of callbacks doesn't match the number of signals ``ValueError`` will be raised.
```

tests/test_wait_signal.py
nicoddemus: I usually don't mind some whitespace changes in the middle of the PR if they don't get in the way of reviewing, so I think the PR as regarding that. :grin:

tests/test_wait_signal.py
```diff
@@ -371,8 +371,269 @@ def test_connected_signal(self, qtbot, signaller):
         assert blocker.args == ['foo', 2342]
 
 
-class TestAssertNotEmitted:
+def test_signal_identity(signaller):
+    """
```
nicoddemus: Excellent docs! :+1:

pytestqt/wait_signal.py
```diff
@@ -123,6 +122,9 @@ def _quit_loop_by_signal(self, *args):
         """
         quits the event loop and marks that we finished because of a signal.
         """
+        if self.callback:
+            if not self.callback(*args):
+                return  # parameter check did not pass
```
nicoddemus: I wonder if we should store the received args, so we can provide a better error message? 

Consider:

```python
def is_disconnected(text):
    return text == 'END'

with qtbot.waitSignal(conn.text_received, callback=is_disconnected):
    conn.request_disconnect()   
```

`conn.text_received` is emitted with `['WAIT_CLIENTS', 'PROCESSING']` and eventually times out.

In this situation all the user will see is a `TimeoutError` and won't have any hints at what might have gone wrong. I suppose the user would then have to add `print` statements at `is_disconnected` to try to figure out the problem.

I think if we store the received arguments when there is a callback, we can reuse that later in the exception message:

```
TimeoutError:  `text_received` emitted with ['WAIT_CLIENTS', 'PROCESSING'] 
but did not satisfy the `is_disconnected` callback.
```

That idea also implies that the same arguments should be available for inspection later in case `raising` is `False`, probably as a list of `*args`: 

```python
blocker.all_args == ['WAIT_CLIENTS', 'PROCESSING']
```

nicoddemus:
@MShekow first let me say thank you for your extremely well written PR!

I reviewed it partially, but I'm afraid is very late and I must get some sleep now. I will continue reviewing it tomorrow.

pytestqt/qtbot.py
MShekow: So, here is a suggestion:
```
:param list check_param_callbacks: list of callables that compare the provided signal parameter(s) to some expected parameter. Each callable has to match the signature of the corresponding signal in ``signals`` (just like a slot function would) and return ``True`` if parameters match, ``False`` otherwise. Instead of a specific callable, ``None`` can be provided, to disable parameter checking for the corresponding signal. If the number of callbacks doesn't match the number of signals ``ValueError`` will be raised.
```

pytestqt/wait_signal.py
MShekow: Sounds good.
Another question is: should this also be implemented for `MultiSignalBlocker`? It would generally be interesting to get a string-dump (in human readable form) of all recorded signals and their parameters, including the points of time where things went wrong (e.g. in case of _strict_ order when the order was violated). Similarly, if `raising == False`, `blocker.recorded_signals` (or something similar) could contain a list of recorded signals together with their parameters.

The-Compiler:
@nicoddemus can you rerun AppVeyor now that #142 is in? (Also, it'd be nice if this plugin was under the pytestbot account on AppVeyor too..)

nicoddemus:
Don't we have to merge with master to get your changes into this pr anyway? This would cause a new build to be triggered anyway. 

About the pytestbot user, I agree, will see how to do it.

The-Compiler:
As far as I know, PRs on AppVeyor/Travis are always built against the latest master (which is also why they don't build when the PR is unmergeable)

nicoddemus:
> As far as I know, PRs on AppVeyor/Travis are always built against the latest master (which is also why they don't build when the PR is unmergeable)

Oh really didn't notice that, but it makes sense! :wink: 

Just triggered a build, let's see how it goes. I'll look later how to move ownership of the project to `pytestbot`.

nicoddemus:
It broke only the `docs` environment now. :grin:

MShekow:
Hi. Are there any other remarks? I'll be going on vacation for the rest of July soon. Before that, I'll have time to fix all the rather "cosmetic" issues ;). There were 2 larger issues mentioned which I cannot complete before that:
1) Argument-capturing for `SignalBlocker` (`all_args`) and `MultiSignalBlocker` (recording the signals and their args in order)
2) More descriptive error message in `SignalTimeoutError`
The 2 points together are a couple of hours of work, given that I'd need to write tests for these new features, too.

nicoddemus:
Hi @MShekow, sorry for the delay.

If you are going on vacation soon, I would rather we merge it sooner and work on improving the error message in `2.1`. @The-Compiler what do you think?

nicoddemus:
@MShekow if you could go ahead and fix the cosmetic issues it would be great.

The-Compiler:
Sorry for the late answer - I was busy with Europython and a big mail-backlog after that, so I postponed everything which was a bit more work :laughing: 

@nicoddemus I'm guessing @MShekow is in holidays by now - what cosmetic issues do you mean? I can see "docstring of pytestqt.qtbot.QtBot.waitSignal:36: WARNING: Inline emphasis start-string without end-string." on Appveyor, but it seems they fixed the other issues in 2e214195ee2872fa8fe5508b07f3cb4a9a310422.

The PR looks good otherwise (though I don't really understand the order handling and tests 100% yet :laughing:).

I'm wondering if we should make `order="strict"` the default though, as we're already breaking stuff with v2.0 anyways.

nicoddemus:
I will try to merge this tonight together with #130 and make the 2.0 release then.

The-Compiler:
@nicoddemus What do you think about my comment regarding the default ordering above?

nicoddemus:
> I'm wondering if we should make order="strict" the default though, as we're already breaking stuff with v2.0 anyways.

Oh forgot to comment: I agree with using `strict` as default,  but this is a new feature so we won't be breaking anything, right?

The-Compiler:
Nope - `MultiSignalBlocker` already existed before and didn't care in which order the signals were emitted. If we turn on the order checking by default now, that *is* a breaking change.

nicoddemus:
Oh right, sorry.

Hmm not sure TBH, it seems strict ordering only makes sense for signals coming from the same object. For multiple signals from different objects like the [example code in the docs](http://pytest-qt.readthedocs.io/en/1.11.0/signals.html#waitsignals) it doesn't make sense and is even error prone.

The-Compiler:
Hmm, true. I reviewed my usage of `waitSignals` and it has many instances where the order is fixed but I don't actually care (e.g. a `Question` object in a prompt emitting `answered`, `completed` and `answered_no`), and some instances where I care (e.g. `[reply.metaDataChanged, reply.readyRead, reply.finished]`). Let's keep it.

nicoddemus:
Merged. Thanks @MShekow again for the PR!

The-Compiler:
So... I'm a bit late, but I just realised `check_params_cb` and `order` aren't mentioned in the [signal docs](http://pytest-qt.readthedocs.io/en/2.0.0/signals.html) at all, and `order` isn't even in the changelog.

nicoddemus:
Thanks, completely missed that. Good thing the documentation is live, so it can be fixed on `master`. Created #147 for it. :grin:

PR #142, opened 2016-07-15
appveyor: Fix registry fixing

The-Compiler:
It's not needed anymore as no alternative Python installations exist,
and it fails now because there's only a Python35 key.

Should fix the issue in #141.

Opening a WIP PR to check if 35 fixing is needed...


The-Compiler:
Okay, seems like it's still needed for Python 3.5 - ready to merge now!

nicoddemus:
Thanks! :grin:

PR #150, opened 2016-08-02
Remove unreachable code

nicoddemus:


pytestqt/qtbot.py
```diff
@@ -422,7 +422,6 @@ def timed_out():
                     return
                 else:
                     assert not timed_out(), 'waitUntil timed out in %s miliseconds' % timeout
```
The-Compiler: Seems like it would be reached if this assert passed?

nicoddemus:
Argh you are right, nevermind; that wait is for 10 **ms**, not 10 **seconds**. :sweat_smile:

Issue #148, opened 2016-08-01
version 2.0.0 breaks pyqt4 api v2


montefra:
Running tests via tox with pytest-qt==2.0.0 with pyqt4 and api v2 gives me:

> py27 runtests: commands[1] | py.test
> INTERNALERROR> Traceback (most recent call last):
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/_pytest/main.py", line 90, in wrap_session
> INTERNALERROR>     config._do_configure()
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/_pytest/config.py", line 839, in _do_configure
> INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 729, in call_historic
> INTERNALERROR>     self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 338, in _hookexec
> INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 333, in <lambda>
> INTERNALERROR>     _MultiCall(methods, kwargs, hook.spec_opts).execute()
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/_pytest/vendored_packages/pluggy.py", line 596, in execute
> INTERNALERROR>     res = hook_impl.function(*args)
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/pytestqt/plugin.py", line 167, in pytest_configure
> INTERNALERROR>     qt_api.set_qt_api(config.getini('qt_api'))
> INTERNALERROR>   File "/.../.tox/py27/local/lib/python2.7/site-packages/pytestqt/qt_compat.py", line 74, in set_qt_api
> INTERNALERROR>     sip.setapi("QDate", 2)
> INTERNALERROR> ValueError: API 'QDate' has already been set to version 1
> ERROR: InvocationError: '/.../.tox/py27/bin/py.test'

Version 1.11.0 runs correctly.

[...]

PR #152, opened 2016-08-05
Update documentation on ``pyqt4v2``

montefra:
I have a tentative update of the documentation according to what we have discussed in pytest-dev/pytest-qt#148.

I'm not very satisfied by it. I might change some of it in the next days. Comments are welcome

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -79,9 +79,13 @@ activate a new fresh environment and execute::
 
     git clone https://github.com/pytest-dev/pytest-qt.git
     cd pytest-qt
-    python setup.py develop
+    pip install -e .  # or python setup.py develop
     pip install pyside # or pyqt4/pyqt5
 
+If you also intend to build the documentation locally, you can make sure to have
+all the needed dependences executing::
+
+    pip install -e .[doc]
 
 .. _virtualenv: https://virtualenv.readthedocs.io/
 
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -51,6 +51,9 @@ Alternatively, you can set the ``PYTEST_QT_API`` environment variable to the
 same values described above (the environment variable wins over the
 configuration if both are set).
 
+From ``pytest-qt`` version 2 the behaviour of ``pyqt4v2`` has changed, as
+explained in :doc:`note_pyqt4v2`.
+
 .. _version 2: http://pyqt.sourceforge.net/Docs/PyQt4/incompatible_apis.html
 
 Installation
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -40,19 +40,39 @@ this order:
 
 To force a particular API, set the configuration variable ``qt_api`` in your ``pytest.ini`` file to
 ``pyqt5``, ``pyside``, ``pyqt4`` or ``pyqt4v2``. ``pyqt4v2`` sets the ``PyQt4``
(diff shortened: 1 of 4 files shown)


docs/intro.rst
```diff
@@ -40,19 +40,39 @@ this order:
...
     [pytest]
     qt_api=pyqt5
 
-
 Alternatively, you can set the ``PYTEST_QT_API`` environment
 variable to the same values described above (the environment variable wins over the configuration
 if both are set).
 
-.. _version2: http://pyqt.sourceforge.net/Docs/PyQt4/incompatible_apis.html
+Starting with ``pytest-qt`` version 2.0, ``PyQt`` or ``PySide`` are loaded after
```
nicoddemus: Probably this is too much information to be into the "introduction" part of the docs. Perhaps we can add an entry "A note about pyqt4v2", similar to [A note about QApplication.exit()](http://pytest-qt.readthedocs.io/en/latest/app_exit.html)?

docs/intro.rst
```diff
@@ -40,19 +40,39 @@ this order:
...
+the ``conftest.py`` files but before all the test modules. This can lead to some
+unexpected behaviour if ``pyqt4v2``.
```
nicoddemus: Probably missing a "if pyqt4v2 is set" or something like this.

nicoddemus:
Thanks for the follow up @montefra!

docs/intro.rst
montefra: right. Fixed

docs/intro.rst
```diff
@@ -40,19 +40,39 @@ this order:
...
     qt_api=pyqt5
 
-
 Alternatively, you can set the ``PYTEST_QT_API`` environment
 variable to the same values described above (the environment variable wins over the configuration
 if both are set).
 
-.. _version2: http://pyqt.sourceforge.net/Docs/PyQt4/incompatible_apis.html
+Starting with ``pytest-qt`` version 2.0, ``PyQt`` or ``PySide`` are loaded after
+the ``conftest.py`` files but before all the test modules. This can lead to some
```
montefra: Much better than mine version. A new commit is arriving

montefra:
@nicoddemus I've updated the info according to your suggestions.

----

on a slightly related note: would you mind if I add ``extras_require={'doc': ['sphinx', 'sphinx_rtd_theme']},`` in the ``setup`` call and update the Development information to use ``pip install -e .`` or ``pip install -e .[doc]`` , instead of ``python setup.py develop``?
I found myself using a couple of different venvs and having to run ``make html``,  ``pip install sphinx``, ``make html`` and then ``pip install sphinx_rtd_theme`` to be able to build the documentation locally.

nicoddemus:
> on a slightly related note: would you mind if I add extras_require={'doc': ['sphinx', 'sphinx_rtd_theme']}, in the setup call and update the Development information to use pip install -e . or pip install -e .[doc] , instead of python setup.py develop?

Not at all, seems like a much better approach than the current one. :grin:

montefra:
can I do it here or would you prefer a new branch?

nicoddemus:
Feel free to do it here, just make sure to use a separate commit. :+1:

nicoddemus:
Excellent, many thanks @montefra!

montefra:
Glad to contribute back to a tool I use regularly.
Thank you for accepting the PR.

Issue #151, opened 2016-08-03
Improve waitSignal/waitSignals TimeoutError messages and catch arguments


MShekow:
Hi. I've started thinking about implementing these 2 features (as they are closely related). Originally @nicoddemus [suggested](https://github.com/pytest-dev/pytest-qt/pull/141/#discussion_r70916993) that `SignalBlocker` could show the following message:
```
TimeoutError:  `text_received` emitted with ['WAIT_CLIENTS', 'PROCESSING'] 
but did not satisfy the `is_disconnected` callback.
```
I investigated a bit and it seems that it's not possible to get the actual name of the signal (here: `text_received`) within `SignalBlocker`. The reason is that Signal objects don't have a `__name__` attribute. PyQt4/5 signals have an internal `signal` field which contains the name of the signal itself, prefixed with a `'2'` (don't ask me why, I'll ask in the pyqt mailing list), whereas PySide signals have no attributes at all, their `str()` representation is just something like `<PySide.QtCore.SignalInstance object at 0x05D82880>`.

I'm not sure how to continue now. For `SignalBlocker` it's not much of an issue, as the `wait()` call of the blocker expects just one signal anyway, and the code line number in the stack trace of a failed test will allow the test engineer to figure out which signal exactly failed.

However, I'd also like to add this feature to `MultiSignalBlocker`. Here it would be nice to have a human-readable form of all caught signals, including their parameters. From what we currently know, it wouldn't be possible to get the name of the emitted signals.

After a bit more digging, I found that there could be a solution, if we can get `_AbstractSignalBlocker` to inherit from `QObject`. Because then we can get the signal by getting the signal's sender and iterating over its attributes, as follows:
```
# the following code (Python 3) works when placed inside SignalBlocker._quit_loop_by_signal() and
# you manually change that _AbstractSignalBlocker inherits from QObject of some Qt framework
def get_signal_name():
    sender = self.sender()
    possible_signal_names = [possible_signal_name for possible_signal_name in sender.__dict__]
    if not possible_signal_names:
        # e.g. PyQt5 has an empty __dict__ for some reason...
        possible_signal_names = dir(self.sender())
    for possible_signal_name in possible_signal_names:
        possible_signal = getattr(sender, possible_signal_name, None)
        if possible_signal and str(self._signals[0]) == str(possible_signal):
            return possible_signal_name
    return "UNKNOWN SIGNAL"
```
[...]

PR #153, opened 2016-08-12
Support for capturing signals and args and improved SignalTimeoutError messages

MShekow:
``waitSignal()``: added support for capturing arguments when callback was not satisfied (into ``blocker.all_args``) and improved the message of ``SignalTimeoutError`` to include the name of the signal, non-matching signal parameters and name of the callback.

``waitSignals()``: Added capturing signals and their arguments (see ``all_signals_and_args``). Cleaned up large parts of the code of ``MultiSignalBlocker`` for better readability. Improved the message of ``SignalTimeoutError``. Updated documentation. Added tests. See #151

### DIFF
--- a/docs/signals.rst
+++ b/docs/signals.rst
@@ -114,6 +114,16 @@ of the blocker:
 Signals without arguments will set ``args`` to an empty list. If the time out
 is reached instead, ``args`` will be ``None``.
 
+Getting all arguments of non-matching arguments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.0
+
+When using the ``check_params_cb`` parameter, it may happen that the provided signal is received multiple times with
+different parameter values, which may or may not match the requirements of the callback.
+``all_args`` then contains the list of signal parameters (as tuple) in the order they were received.
+
+
 waitSignals
 -----------
 
@@ -174,7 +184,7 @@ evaluation takes place).
 
 
 order parameter
-^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^
 
 .. versionadded:: 2.0
 
@@ -193,6 +203,17 @@ the signals ``[a, b]`` but the sender emitted signals ``[a, a, b]``.
 A third option is to set ``order="simple"`` which is like "strict", but signals may be emitted
 in-between the provided ones, e.g. if the expected signals are ``[a, b, c]`` and the sender 
 actually emits ``[a, a, b, a, c]``, the test completes successfully (it would fail with ``order="strict"``).
+
+Getting emitted signals and arguments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.1
+
+To determine which of the expected signals were emitted during a ``wait()`` you can use
+``blocker.all_signals_and_args`` which contains a list of
+:class:`wait_signal.SignalAndArgs <SignalAndArgs>` ``namedtuple`` objects, indicating the signals (and their arguments)
+in the order they were received.
+
 
 Making sure a given signal is not emitted
 -----------------------------------------
--- a/docs/signals.rst
+++ b/docs/signals.rst
@@ -211,7 +211,7 @@ Getting emitted signals and arguments
 
 To determine which of the expected signals were emitted during a ``wait()`` you can use
 ``blocker.all_signals_and_args`` which contains a list of
-:class:`wait_signal.SignalAndArgs <SignalAndArgs>` ``namedtuple`` objects, indicating the signals (and their arguments)
+:class:`wait_signal.SignalAndArgs <SignalAndArgs>` objects, indicating the signals (and their arguments)
 in the order they were received.
 
 
--- a/docs/signals.rst
+++ b/docs/signals.rst
@@ -117,7 +117,7 @@ is reached instead, ``args`` will be ``None``.
 Getting all arguments of non-matching arguments
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. versionadded:: 2.0
+.. versionadded:: 2.1
 
 When using the ``check_params_cb`` parameter, it may happen that the provided signal is received multiple times with
 different parameter values, which may or may not match the requirements of the callback.
--- a/pytestqt/wait_signal.py
+++ b/pytestqt/wait_signal.py
@@ -20,6 +20,7 @@ class _AbstractSignalBlocker(object):
         self.timeout = timeout
         self.signal_triggered = False
         self.raising = raising
+        self._signals = None  # will be initialized by inheriting implementations
         if timeout is None:
             self._timer = None
         else:
@@ -44,8 +45,8 @@ class _AbstractSignalBlocker(object):
             self._timer.start()
         self._loop.exec_()
         if not self.signal_triggered and self.raising:
-            raise SignalTimeoutError("Didn't get signal after %sms." %
-                                     self.timeout)
+            # raise SignalTimeoutError("Didn't get signal after %sms." % self.timeout)
+            raise SignalTimeoutError(self.get_timeout_error_message())
 
     def _quit_loop_by_timeout(self):
         try:
@@ -58,6 +59,64 @@ class _AbstractSignalBlocker(object):
             _silent_disconnect(self._timer.timeout, self._quit_loop_by_timeout)
             self._timer.stop()
             self._timer = None
+
+    def get_timeout_error_message(self):
+        pass
+
+    def _extract_pyqt_signal_name(self, potential_pyqt_signal):
+        signal_name = potential_pyqt_signal.signal  # type: str
+        if type(signal_name) != str:
+            raise TypeError("Invalid 'signal' attribute in {}. "
+                            "Expected str but got {}".format(signal_name, type(signal_name)))
+        # strip magic number "2" that PyQt prepends to the signal names
+        if signal_name.startswith("2"):
+            signal_name = signal_name.lstrip('2')
+        return signal_name
+
+    def _extract_signal_from_signal_tuple(self, potential_signal_tuple):
+        if type(potential_signal_tuple) is tuple:
+            if len(potential_signal_tuple) != 2:
+                raise AssertionError("Signal tuple must have length of 2 (first element is the signal, "
+                                     "the second element is the signal's name).")
+            signal_tuple = potential_signal_tuple
+            signal_name = signal_tuple[1]
+            if type(signal_tuple) != str or not signal_name:
+                raise TypeError("Invalid type for user-provided signal name, "
+                                "expected str but got {}".format(type(signal_name)))
+            return signal_name
+        return ""
+
+    def get_signal_name(self, potential_signal_tuple):
+        """
+        Attempts to extract the signal's name. If the user provided the signal name as 2nd value of the tuple, this
+        name has preference. Bad values cause a ``ValueError``.
+        Otherwise it attempts to get the signal from the ``signal`` attribute of ``signal`` (which only exists for
+        PyQt signals).
+        :returns: str name of the signal, an empty string if no signal name can be determined, or raises an error
+            in case the user provided an invalid signal name manually
+        """
+        signal_name = self._extract_signal_from_signal_tuple(potential_signal_tuple)
+
+        if not signal_name:
+            try:
+                signal_name = self._extract_pyqt_signal_name(potential_signal_tuple)
+            except AttributeError:
+                # not a PyQt signal
+                # -> no signal name could be determined
+                signal_name = ""
+
+        return signal_name
+
+    def get_callback_name(self, callback):
+        """Attempts to extract the name of the callback. Returns empty string in case of failure."""
+        try:
+            name = callback.__name__
+        except AttributeError:
+            try:
+                name = callback.func.__name__  # e.g. for callbacks wrapped with functools.partial()
+            except AttributeError:
+                name = ""
+        return name
 
     def __enter__(self):
         return self
@@ -103,7 +162,9 @@ class SignalBlocker(_AbstractSignalBlocker):
         super(SignalBlocker, self).__init__(timeout, raising=raising)
         self._signals = []
         self.args = None
+        self.all_args = []
         self.check_params_callback = check_params_cb
+        self.signal_name = ""
 
     def connect(self, signal):
         """
@@ -115,6 +176,7 @@ class SignalBlocker(_AbstractSignalBlocker):
 
         :param signal: QtCore.Signal
         """
+        self.signal_name = self.get_signal_name(potential_signal_tuple=signal)
         signal.connect(self._quit_loop_by_signal)
         self._signals.append(signal)
 
@@ -123,6 +185,7 @@ class SignalBlocker(_AbstractSignalBlocker):
         quits the event loop and marks that we finished because of a signal.
         """
         if self.check_params_callback:
+            self.all_args.append(args)
             if not self.check_params_callback(*args):
                 return  # parameter check did not pass
         try:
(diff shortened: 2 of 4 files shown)


pytestqt/wait_signal.py
```diff
@@ -44,8 +48,8 @@ def wait(self):
             self._timer.start()
         self._loop.exec_()
         if not self.signal_triggered and self.raising:
-            raise SignalTimeoutError("Didn't get signal after %sms." %
-                                     self.timeout)
+            # raise SignalTimeoutError("Didn't get signal after %sms." % self.timeout)
```
The-Compiler: I guess you can get rid of this line?

pytestqt/wait_signal.py
```diff
@@ -54,11 +58,77 @@ def _quit_loop_by_timeout(self):
...
             _silent_disconnect(self._timer.timeout, self._quit_loop_by_timeout)
             self._timer.stop()
             self._timer = None
 
+    def get_timeout_error_message(self):
+        pass
+
+    def _extract_pyqt_signal_name(self, potential_pyqt_signal):
+        signal_name = potential_pyqt_signal.signal  # type: str
+        if type(signal_name) != str:
```
The-Compiler: `if isinstance(signal_name, str)` would be more idiomatic

pytestqt/wait_signal.py
```diff
@@ -54,11 +58,77 @@ def _quit_loop_by_timeout(self):
...
+            raise TypeError("Invalid 'signal' attribute in {}. "
+                            "Expected str but got {}".format(signal_name, type(signal_name)))
+        # strip magic number "2" that PyQt prepends to the signal names
+        if signal_name.startswith("2"):
```
The-Compiler: No need for the check, `lstrip` just won't strip anything if it's not there.

pytestqt/wait_signal.py
```diff
@@ -54,11 +58,77 @@ def _quit_loop_by_timeout(self):
...
+            signal_name = signal_name.lstrip('2')
+        return signal_name
+
+    def _extract_signal_from_signal_tuple(self, potential_signal_tuple):
+        if type(potential_signal_tuple) is tuple:
```
The-Compiler: `isinstance` as above

pytestqt/wait_signal.py
```diff
@@ -54,11 +58,77 @@ def _quit_loop_by_timeout(self):
...
+            if len(potential_signal_tuple) != 2:
+                raise AssertionError("Signal tuple must have length of 2 (first element is the signal, "
+                                     "the second element is the signal's name).")
+            signal_tuple = potential_signal_tuple
+            signal_name = signal_tuple[1]
+            if type(signal_name) != str or not signal_name:
```
The-Compiler: `isinstance` as above

pytestqt/wait_signal.py
```diff
@@ -54,11 +58,77 @@ def _quit_loop_by_timeout(self):
...
+                            "Expected str but got {}".format(signal_name, type(signal_name)))
+        # strip magic number "2" that PyQt prepends to the signal names
+        if signal_name.startswith("2"):
+            signal_name = signal_name.lstrip('2')
+        return signal_name
+
+    def _extract_signal_from_signal_tuple(self, potential_signal_tuple):
+        if type(potential_signal_tuple) is tuple:
+            if len(potential_signal_tuple) != 2:
+                raise AssertionError("Signal tuple must have length of 2 (first element is the signal, "
```
The-Compiler: `ValueError` maybe?

pytestqt/wait_signal.py
```diff
@@ -54,11 +58,77 @@ def _quit_loop_by_timeout(self):
...
+        except AttributeError:
+            try:
+                name = callback.func.__name__  # e.g. for callbacks wrapped with functools.partial()
+            except AttributeError:
+                name = ""
+        return name
+
+    @staticmethod
+    def get_signal_from_potential_signal_tuple(signal_tuple):
+        if type(signal_tuple) is tuple:
```
The-Compiler: `isinstance` as above

pytestqt/wait_signal.py
```diff
@@ -138,6 +213,63 @@ def _cleanup(self):
...
+
+
+SignalAndArgs = namedtuple("SignalAndArgs", ["signal_name", "args"])
+
+
+def _get_readable_signal_with_optional_args(self):
+    if self.args:
+        args_as_string = []
+        for arg in self.args:
+            if type(arg) is str:
```
The-Compiler: `isinstance` as above

pytestqt/wait_signal.py
```diff
@@ -138,6 +213,63 @@ def _cleanup(self):
...
+                args_as_string.append("'" + str(arg) + "'")
+            else:
+                args_as_string.append(str(arg))
```
The-Compiler: I guess it'd be better to just call `repr()` on all arguments?

pytestqt/wait_signal.py
```diff
@@ -138,6 +213,63 @@ def _cleanup(self):
...
+        args_as_list_string = ", ".join(args_as_string) if len(args_as_string) > 1 else args_as_string[0]
```
The-Compiler: `", ".join(["foo"])` -> `"foo"`, so no need for that check

pytestqt/wait_signal.py
```diff
@@ -138,6 +213,63 @@ def _cleanup(self):
...
+        args = "({})".format(args_as_list_string)
+    else:
+        args = ""
+
+    # remove signal parameter signature, e.g. turn "some_signal(QString,int)" to "some_signal", because we're adding
+    # the actual parameters anyways
+    signal_name = self.signal_name
+    if '(' in signal_name:
+        signal_name = signal_name[:signal_name.index('(')]
```
The-Compiler: You could just do `signal_name = signal_name.partition('(')[0]` unconditionally

pytestqt/wait_signal.py
```diff
@@ -138,6 +213,63 @@ def _cleanup(self):
...
+
+    return signal_name + args
+
+
+SignalAndArgs.__str__ = _get_readable_signal_with_optional_args
+SignalAndArgs.__repr__ = _get_readable_signal_with_optional_args
```
The-Compiler: Might make more sense to have a custom class instead of abusing a `namedtuple` and then monkeypatching stuff on it?

tests/test_wait_signal.py
```diff
@@ -640,6 +641,332 @@ def test_signals_and_callbacks_length_mismatch(self, qtbot, signaller):
...
+        """
+        Tests that all_signals_and_args contains the emitted signals. No callbacks for arg-evaluation are provided. The
+        signals are emitted out of order, causing a timeout.
+        """
+        signals = get_mixed_signals_with_guaranteed_name(signaller)
+        with qtbot.waitSignals(signals=signals, timeout=200, check_params_cbs=None, order="simple",
+                               raising=False) as blocker:
+            signaller.signal_args.emit('1', 1)
+            signaller.signal.emit()
+        assert blocker.signal_triggered is False
```
The-Compiler: just use `assert not blocker.signal_triggered` (same for other places where you use `is True` and `is False`)

tests/test_wait_signal.py
```diff
@@ -640,6 +641,332 @@ def test_signals_and_callbacks_length_mismatch(self, qtbot, signaller):
...
+        def arg_validator(str_param, int_param):
+            return str_param == "1337" and int_param == 1337
+
+        with pytest.raises(SignalTimeoutError) as excinfo:
+            with qtbot.waitSignal(signal=signal, timeout=200, check_params_cb=arg_validator, raising=True):
+                signaller.signal_args.emit('1', 1)
+                signaller.signal_args.emit('2', 2)
+        ex_msg = TestWaitSignalsSignalTimeoutErrorMessage.get_exception_message(excinfo)
+        assert ex_msg == "Signal signal_args(QString,int) emitted with parameters [('1', 1), ('2', 2)] " \
+                         "within 200 ms, but did not satisfy the arg_validator callback"
```
The-Compiler: You can just use parens for line continuation instead of using `\` - I think `\` is too brittle, e.g. accidentally inserting a space after it suddenly won't do what you expect anymore.

(Same for other message checks in this file)

The-Compiler:
Thanks for your great work on this, looking forward to it!

I added some style comments, and plan to take another closer look when I feel more awake :laughing: 

Some other things to note:

- The coverage should stay at 100% - can you investigate what's not covered yet?
- The tests seem to fail with Python 2 and PyQt4/PySide - can you investigate?

MShekow:
Hi. I'm having trouble fixing the bugs. The most problematic issue is unicode strings. Python 2 and 3 seem to work very different (I don't personally care for legacy software, so I never use Python 2). Python 2 breaks my tests because it converts ``str( ('somestring',) )`` to ``"(u'somestring',)"`` (i.e. it adds the ``u`` which I don't want there.
What I can do is to have a ``if type(somestring) is unicode`` check when iterating over the tuple's elements (and convert them with ``str()`` in this case), but that only works in Python 2 (Python 3 doesn't recognize ``unicode``). So, what do I do now? Do I really have to do Python version checking in my code to fix this?

pytestqt/wait_signal.py
MShekow: I'm fine with replacing ``type()`` with ``isinstance()``. 
Regarding ``assert`` you're kind of disagreeing with yourself (see https://github.com/pytest-dev/pytest-qt/pull/141#discussion_r70802896 ), aren't you?

The-Compiler: Hmm, fair enough, sorry :laughing: Good point former-me made there, so let's keep it :wink:

The-Compiler:
Why don't you want the `u'...'` there? If a parameter is an unicode string, I'd surely want to see that fact, just like I want to have `'` around strings. I'd suggest just adjusting the test accordingly (with a version check probably).

MShekow:
Well, I didn't want the u there due to the tests. I'll follow your suggestion and adapt the tests instead.

MShekow:
So, I've fixed all those things you mentioned and got the tests working again.
Regarding the coverage: I can add tests for those lines the tests do not currently cover. These are (see here for more details: https://coveralls.io/builds/7422331/source?filename=pytestqt%2Fwait_signal.py ):
1) waitSignals() code that raises ``TypeError`` or ``ValueError`` when user doesn't provide valid parameters
2) Line 120: no name for a provided callback could be determined. I'm not sure how to cover that. Maybe callable wrapped inside a ``functools.partial`` wrapped inside a ``functools partial``?

docs/signals.rst
```diff
@@ -194,6 +204,17 @@ A third option is to set ``order="simple"`` which is like "strict", but signals
 in-between the provided ones, e.g. if the expected signals are ``[a, b, c]`` and the sender 
 actually emits ``[a, a, b, a, c]``, the test completes successfully (it would fail with ``order="strict"``).
 
+Getting emitted signals and arguments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.1
```
MShekow: Is this guess for the next version actually correct?

MShekow:
I've done the necessary changes. I don't know how to further increase the coverage, tbh.

pytestqt/wait_signal.py
```diff
@@ -54,11 +57,76 @@ def _quit_loop_by_timeout(self):
...
     def _cleanup(self):
+        # store timeout message before the data to construct it is lost
+        self._timeout_message = self.get_timeout_error_message()
         if self._timer is not None:
             _silent_disconnect(self._timer.timeout, self._quit_loop_by_timeout)
             self._timer.stop()
             self._timer = None
 
+    def get_timeout_error_message(self):
+        pass
```
The-Compiler: Since this must be overridden in subclasses, you should do `raise NotImplementedError` here ([docs](https://docs.python.org/3.5/library/exceptions.html#NotImplementedError)) - that should also mean no coverage is required for it, and if coverage.py doesn't do so automatically, we can [tell it to](https://coverage.readthedocs.io/en/default/excluding.html#advanced-exclusion).

MShekow: I've changed it in the code, but coveralls still complains about that line not being covered. So, should I use a pragma? I don't see any configuration files in place yet.

nicoddemus:
> I've done the necessary changes. I don't know how to further increase the coverage, tbh.

For this case specifically I would say it is fair to add a `# pragma: no cover` on that line. :+1:

docs/signals.rst
```diff
@@ -114,6 +114,16 @@ of the blocker:
 Signals without arguments will set ``args`` to an empty list. If the time out
 is reached instead, ``args`` will be ``None``.
 
+Getting all arguments of non-matching arguments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.0
```
nicoddemus: This should be 2.1?

docs/signals.rst
nicoddemus: Yes! :grin:

pytestqt/wait_signal.py
```diff
@@ -54,11 +57,77 @@ def _quit_loop_by_timeout(self):
...
 
     def _cleanup(self):
+        # store timeout message before the data to construct it is lost
+        self._timeout_message = self.get_timeout_error_message()
         if self._timer is not None:
             _silent_disconnect(self._timer.timeout, self._quit_loop_by_timeout)
             self._timer.stop()
             self._timer = None
 
+    def get_timeout_error_message(self):
```
nicoddemus: I'm not sure this should be part of the public API, so it should have a `_` prefix (It is OK for subclasses to access members prefixed with  `_`).

pytestqt/wait_signal.py
```diff
@@ -54,11 +57,77 @@ def _quit_loop_by_timeout(self):
...
+        """Subclasses have to implement this, returning an appropriate error message for a SignalTimeoutError."""
+        raise NotImplementedError
+
+    def _extract_pyqt_signal_name(self, potential_pyqt_signal):
```
nicoddemus: The functions below, which extract a string/name from signal objects, could all live in a `signals.py` module (or something like that) as normal functions instead of methods. It would make it simpler to test each one of them as unit-tests.

pytestqt/wait_signal.py
```diff
@@ -154,107 +282,226 @@ class MultiSignalBlocker(_AbstractSignalBlocker):
...
+                successful_index = self._get_first_matching_index(unique_signal, *args)
                 self._signals_emitted[successful_index] = True
-        elif self.order == "simple":
-            potential_indices = self._get_unemitted_signal_indices(signal)
-            if potential_indices:
-                if self._signal_expected_index == potential_indices[0]:
-                    if self._check_callback(self._signal_expected_index, *args):
-                        self._signals_emitted[self._signal_expected_index] = True
-                        self._signal_expected_index += 1
+            except:  # none found
```
nicoddemus: Better to be explicit about which exceptions you plan to catch here... a bare except like this is considered somewhat dangerous because it will capture absolutely any exception

nicoddemus:
Hey @MShekow, excellent documentation on the tests and code!

Other than adding the comment to skip the missing coverage, my suggestions are just that, suggestions, so feel free to not implement them if you don't really have the time as I'm guessing you have spent quite a few hours on this already.

All your work here is much appreciated!

pytestqt/wait_signal.py
The-Compiler: This is a must-have IMHO - here you'd even catch things like a typo (or if you're unlucky, the user pressing ctrl-c) in the `try` block :wink:

pytestqt/wait_signal.py
```diff
@@ -54,11 +57,77 @@ def _quit_loop_by_timeout(self):
...
+    def _extract_signal_from_signal_tuple(self, potential_signal_tuple):
+        if isinstance(potential_signal_tuple, tuple):
+            if len(potential_signal_tuple) != 2:
+                raise ValueError("Signal tuple must have length of 2 (first element is the signal, "
+                                 "the second element is the signal's name).")
+            signal_tuple = potential_signal_tuple
+            signal_name = signal_tuple[1]
+            if not isinstance(signal_name, str) or not signal_name:
+                raise TypeError("Invalid type for user-provided signal name, "
+                                "expected str but got {}".format(type(signal_name)))
```
The-Compiler: If someone accidentally passed an empty string, they'd get "Invalid type for user-provided signal name, expected str but got str", which might be quite confusing.

pytestqt/wait_signal.py
```diff
@@ -138,6 +212,60 @@ def _cleanup(self):
...
+
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return self.__dict__ == other.__dict__
+        else:
+            return False
+
+
+# Returns e.g. "3rd" for 3, or "21st" for 21
+get_ordinal_str = lambda n: "%d%s" % (n, {1: "st", 2: "nd", 3: "rd"}.get(n if n < 20 else n % 10, "th"))
```
The-Compiler: That's clever :laughing:

MShekow: Yes. I found it on the Internet :)

pytestqt/wait_signal.py
MShekow: You're right. I'll fix this, i.e. split it into two ``if``s (each one without any ``or``).

pytestqt/wait_signal.py
MShekow: Agreed. I'll fix it by introducing a new exception called ``NoMatchingIndexFoundError``.

pytestqt/wait_signal.py
MShekow: Although I agree, I suppose the functions can be moved once actually any unit tests are written for them ;).

nicoddemus:
Hey @MShekow,

Really sorry about the delay on this. Just now I found the time to review it and it looks great! Your entire PR is very well written and your test descriptions are really helpful! :+1: 

Long overdue, I'm merging this! I plan to have 2.1 out soon real soon. :grin: 

Also, thanks @The-Compiler for the through review as well. :wink:

MShekow:
My pleasure! Thanks for merging :)

PR #155, opened 2016-08-17
qtbot now has `capture_exceptions` context manager method

### LINKED ISSUE #154: Allow to capture Qt virtual method exceptions in a block
Currently it is possible to do this already, but it requires importing some modules that probably would be better left hidden. I guess allowing code like example below be created using just public `qtbot` interface would be better.

```python
@pytest.mark.qt_no_exception_capture
def test_foo(qtbot):
    from PyQt5.QtCore import pyqtSignal
    from PyQt5.QtWidgets import QWidget

    class FooWidget(QWidget):

        on_event = pyqtSignal()

    widget = FooWidget()
    qtbot.addWidget(widget)

    def raise_on_event():
        raise RuntimeError("ops")

    widget.on_event.connect(raise_on_event)

    from pytestqt.plugin import capture_exceptions
        widget.on_event.emit()

    assert len(exceptions) == 1
    assert exceptions[0][1].message == "ops"
```

I can create the PR myself if you just guide me how you'd prefer this to be implemented.

fogo:
Fixes #154

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,11 @@
+2.1
+---
+
+- ``qtbot`` fixture now can capture Qt virtual method exceptions in a block using
+  ``capture_exceptions`` (`#154`_). Thanks to `@fogo`_ for the PR.
+
+.. _#154: https://github.com/pytest-dev/pytest-qt/issues/154
+
 2.0
 ---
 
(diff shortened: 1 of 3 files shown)


nicoddemus:
Thanks @fogo!

PR #157, opened 2016-08-20
WIP: Add qtbot.waitCallback

The-Compiler:
Some TODO items and questions:
- [ ] Add tests for timeouts
- [ ] Any other tests?
- [ ] Add docs (in `signals.rst`?)
- [ ] Should we re-use `qt_wait_signal_raising`? Or have a different config option? Or rename `qt_wait_signal_raising` to `qt_raising`, use it for both, and have `qt_wait_signal_raising` as deprecated alias?
- [ ] Should we re-use `SignalTimeoutError`? Or have the current `CallbackTimeoutError`? Or rename `SignalTimeoutError` to `TimeoutError`, use it for both, and have `SignalTimeoutError` as deprecated alias?
- [ ] Should I inherit from some `AbstractBlocker` instead of copy-pasting some code to `CallbackBlocker`? I thought it'd make stuff too complicated and a little copy-paste is probably simpler.

### DIFF
--- a/pytestqt/qtbot.py
+++ b/pytestqt/qtbot.py
@@ -2,7 +2,9 @@ import functools
 import contextlib
 import weakref
 
-from pytestqt.wait_signal import SignalBlocker, MultiSignalBlocker, SignalTimeoutError, SignalEmittedSpy
+from pytestqt.wait_signal import (SignalBlocker, MultiSignalBlocker,
+                                  SignalTimeoutError, SignalEmittedSpy,
+                                  CallbackBlocker, CallbackTimeoutError)
 from pytestqt.qt_compat import qt_api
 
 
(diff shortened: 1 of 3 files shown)


nicoddemus:
> Add docs (in signals.rst?)

I agree, but perhaps rename it to `signals_and_callbacks`?

> Should we re-use qt_wait_signal_raising? Or have a different config option? Or rename qt_wait_signal_raising to qt_raising, use it for both, and have qt_wait_signal_raising as deprecated alias?

I think a single option is enough, how about `qt_default_raising` and deprecate `qt_wait_signal_raising`?

> Should we re-use SignalTimeoutError? Or have the current CallbackTimeoutError? Or rename SignalTimeoutError to TimeoutError, use it for both, and have SignalTimeoutError as deprecated alias?

`TimeoutError` and deprecate `SignalTimeoutError` seems simpler, I don't think having a separate exception brings anything useful to the table.

> Should I inherit from some AbstractBlocker instead of copy-pasting some code to CallbackBlocker? I thought it'd make stuff too complicated and a little copy-paste is probably simpler.

Go with your gut feeling, I'm OK with some copy-paste if it makes the code clearer; `AbstractBlocker` and its subclasses seem already a little too complicated as they are.

I don't have time right now to actually review the code, will try to get to it tomorrow. Thanks Florian! :grin:

nicoddemus:
@The-Compiler I'm closing this for now as this has become quite stale... feel free to submit a new PR whenever you get around to implement this. 🤗

PR #159, opened 2016-09-16
Add waitExposed and waitActive methods to QtBot

### LINKED ISSUE #158: waitForWindowShown doesn't handle timeout
Qt's [`qWaitForWindowShown`](https://doc.qt.io/qt-5/qtest-obsolete.html#qWaitForWindowShown) (or [`qWaitForWindowExposed`](https://doc.qt.io/qt-5/qtest.html#qWaitForWindowExposed) which supersedes it since Qt 5.0) has a `timeout` argument, and returns a `bool` saying if the waiting has timed out or not.

pytest-qt's wrapper over it doesn't have the `timeout` argument, discards the return value and doesn't raise an exception either in case there was a timeout.

nicoddemus:
Added `waitForWindowExposed` for PyQt5.

Fixes #158

### DIFF
--- a/pytestqt/qtbot.py
+++ b/pytestqt/qtbot.py
@@ -147,7 +147,7 @@ class QtBot(object):
 
     def waitActive(self, widget, timeout=1000):
         """
-        Context manager that waits for timeout milliseconds or until the window is active.
+        Context manager that waits for ``timeout`` milliseconds or until the window is active.
         If window is not exposed within ``timeout`` milliseconds, raise ``TimeoutError``.
 
         This is mainly useful for asynchronous systems like X11, where a window will be mapped to screen
@@ -159,10 +159,10 @@ class QtBot(object):
                 show_action()
 
         :param QWidget widget:
-            Widget to wait on.
+            Widget to wait for.
 
         :param int|None timeout:
-            How many miliseconds to wait on.
+            How many milliseconds to wait for.
 
         .. note::
             This function is only available in PyQt5, raising a ``RuntimeError`` if called from
@@ -177,7 +177,7 @@ class QtBot(object):
 
     def waitExposed(self, widget, timeout=1000):
         """
-        Context manager that waits for timeout milliseconds or until the window is exposed.
+        Context manager that waits for ``timeout`` milliseconds or until the window is exposed.
         If the window is not exposed within ``timeout`` milliseconds, raise ``TimeoutError``.
 
         This is mainly useful for asynchronous systems like X11, where a window will be mapped to screen
@@ -189,10 +189,10 @@ class QtBot(object):
                 startup()
 
         :param QWidget widget:
-            Widget to wait on.
+            Widget to wait for.
 
         :param int|None timeout:
-            How many miliseconds to wait on.
+            How many milliseconds to wait for.
 
         .. note::
             This function is only available in PyQt5, raising a ``RuntimeError`` if called from
@@ -214,7 +214,7 @@ class QtBot(object):
         :param QWidget widget:
             Widget to wait on.
 
-        .. note:: In ``PyQt5`` this function is considered deprecated in favor of :meth:`waitForWindowExposed`.
+        .. note:: In ``PyQt5`` this function is considered deprecated in favor of :meth:`waitExposed`.
 
         .. note:: This method is also available as ``wait_for_window_shown`` (pep-8 alias)
         """
--- a/pytestqt/qtbot.py
+++ b/pytestqt/qtbot.py
@@ -167,9 +167,13 @@ class QtBot(object):
         .. note::
             This function is only available in PyQt5, raising a ``RuntimeError`` if called from
             ``PyQt4`` or ``PySide``.
+
+        .. note:: This method is also available as ``wait_active`` (pep-8 alias)
         """
         __tracebackhide__ = True
         return _WaitWidgetContextManager('qWaitForWindowActive', 'activated', widget, timeout)
+
+    wait_active = waitActive  # pep-8 alias
 
     def waitExposed(self, widget, timeout=1000):
         """
@@ -193,9 +197,13 @@ class QtBot(object):
         .. note::
             This function is only available in PyQt5, raising a ``RuntimeError`` if called from
             ``PyQt4`` or ``PySide``.
+
+        .. note:: This method is also available as ``wait_exposed`` (pep-8 alias)
         """
         __tracebackhide__ = True
         return _WaitWidgetContextManager('qWaitForWindowExposed', 'exposed', widget, timeout)
+
+    wait_exposed = waitExposed  # pep-8 alias
 
     def waitForWindowShown(self, widget):
         """
--- a/pytestqt/qtbot.py
+++ b/pytestqt/qtbot.py
@@ -1,9 +1,10 @@
-import functools
 import contextlib
+import functools
 import weakref
 
-from pytestqt.wait_signal import SignalBlocker, MultiSignalBlocker, SignalTimeoutError, SignalEmittedSpy
 from pytestqt.qt_compat import qt_api
+from pytestqt.wait_signal import SignalBlocker, MultiSignalBlocker, SignalTimeoutError, SignalEmittedSpy, \
+    SignalEmittedError
 
 
 def _parse_ini_boolean(value):
@@ -26,6 +27,9 @@ class QtBot(object):
     **Widgets**
 
     .. automethod:: addWidget
+    .. automethod:: capture_exceptions
+    .. automethod:: waitActive
+    .. automethod:: waitExposed
     .. automethod:: waitForWindowShown
     .. automethod:: stopForInteraction
     .. automethod:: wait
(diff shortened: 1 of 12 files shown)


nicoddemus:
I plan to release 2.1 after this is merged.

The-Compiler:
I think we still have some things to consider here, ideally before releasing it as they're API changes - unsorted braindump as I don't have much time right now:

- Probably should have [`qWaitForWindowActive`](https://doc.qt.io/qt-5/qtest.html#qWaitForWindowActive) too?
- Why not make it raise on a timeout, with an optional `raising=False`? If we already have a Python wrapper, let's make the API more pythonic and don't force the caller to check a return value, IMHO.
- In the same vein, should it be a context manager? IMHO, this looks better:

```python
with qtbot.waitForWindowExposed(window):
    window.show()
assert window.isVisible()
```

compared to:

```python
window.show()
qtbot.waitForWindowExposed(window)
assert window.isVisible()
```

nicoddemus:
Good point about `waitForWindowActive`... I should also review if there are other additions to `QTest` that should go in.

I agree having a higher level API is nice, but perhaps we should add them as separate methods rather than changing the original functions? I would rather keep the original `QTest` functions "as is" instead of adding additional behavior; it might be a little surprising to users and makes converting C++ examples to Python easier (for example as we did in  the `qmodeltester` fixture).

Following your example, how about those context managers then:

```python
def waitExposed(self, widget, timeout=1000, raising=True):
    pass
def waitActive(self, widget, timeout=1000, raising=True):
    pass
def waitFocus(self, widget, timeout=1000, raising=True):
    pass
```

The last one in particular is tricky to achieve in my experience, specially when executing under `pytest-xdist`.

Also, I think those high-level APIs are only worthwhile doing for PyQt5... let's move forward only. :grin:

The-Compiler:
FWIW we already change the function names by stripping the leading `q` :wink:

I guess adding those would work. No big opinion on backends, I think nobody should use Qt4 nowadays anyways :laughing: 

I also thought about having `qtbot.waitForWindowActive(...)` act as before, but the context manager form raising by default, but I guess that'd be kind of weird too.

nicoddemus:
OK! Thanks for coming back to this. 😁 

I will see if I can tackle this during the week.

nicoddemus:
Hey @The-Compiler,

I implemented `waitForWindowActive` as we discussed.

After implementing the `waitExposed` and `waitActive` context managers as we discussed, I'm not sure they are any better than the actual functions.

Consider:

```python
widget.show()
assert qtbot.qWaitForWindowExposed(widget, timeout)
```

vs:

```python
with qtbot.waitExposed(widget, timeout):
    widget.show()
```

I feel we don't gain much and will bloat the API without compelling reason.

What do you think?

The-Compiler:
The desire for those two things is actually because I make those mistakes a lot. Accidentally using it as a context manager isn't that bad, accidentally thinking it'd be raising is :wink:

The reason I make those mistakes over and over again is because the API is inconsistent - I can use other `qtbot.wait*` methods as a context manager and they'll raise when something is wrong, so I think it's a cognitive burden to have some `wait*` methods behave one way and some `wait*` methods behave differently.

FWIW, I think the cleanest design/API would be to have convenience wrappers as `qtbot.wait*`, and all the `QTest` methods as `qtbot.qtest.*` (or a separate `qtest` fixture even) - but I guess that's 3.0 material :wink:

nicoddemus:
> The reason I make those mistakes over and over again is because the API is inconsistent - I can use other qtbot.wait* methods as a context manager and they'll raise when something is wrong, so I think it's a cognitive burden to have some wait* methods behave one way and some wait* methods behave differently.

Hmm I see, I agree that those are all valid points.

Well, since we are talking about adding two news methods, `waitForWindowExposed` and `waitForWindowActive`, we might decide to not expose them directly and expose a version we find more user-friendly:

```python
def waitActive(self, widget, timeout=1000):
    """
    Raises a TimeoutError if the widget does not show in timeout miliseconds. 
    Calls the QTest.qWaitForWindowActive.
    PyQt5 only.
    """
```

Similarly for `waitExposed`.

1. We might think about adding a `raising=True` parameter, but I'm not sure it is a common need to ensure something is not being activated/exposed which is the only situation I can think of someone wanting to pass `raising=False` and observe the returned value, so I would rather avoid "bloating" the API. It is always simpler to add it later than having to deprecate/remove something.

2. Perhaps those should be context managers, for consistency with the other wait-signal methods?

> FWIW, I think the cleanest design/API would be to have convenience wrappers as qtbot.wait*, and all the QTest methods as qtbot.qtest.* (or a separate qtest fixture even)

I agree it is cleaner, but I'm not sure it is better because it is not as convenient to use. Forcing to use `qtbot.qtest.mouseClick` instead of `qtbot.mouseClick` doesn't seem like a good trade-off in this case. Plus it would break all test suites and would be a little annoying to fix. Don't know, I would have to think about it some more.

Thanks as always for the thoughtful feedback! 😉

The-Compiler:
I agree about all your points! I think they definitely should be context managers for consistency, and I agree we can just default to `raising=True` and not add `raising` until someone has a good reason why it's needed :wink:

nicoddemus:
@The-Compiler I know you are pretty buys, just a ping to let you know this is ready for reviewing. 😁

pytestqt/qtbot.py
```diff
@@ -141,6 +145,66 @@ def addWidget(self, widget):
 
     add_widget = addWidget  # pep-8 alias
 
+    def waitActive(self, widget, timeout=1000):
+        """
+        Context manager that waits for timeout milliseconds or until the window is active.
```
The-Compiler: Shouldn't the `timeout` be monospace?

The-Compiler: Shouldn't the `timeout` be monospace?

pytestqt/qtbot.py
```diff
@@ -141,6 +145,66 @@ def addWidget(self, widget):
...
+        This is mainly useful for asynchronous systems like X11, where a window will be mapped to screen
+        some time after  being asked to show itself on the screen.
+
+        .. code-block:: python
+
+            with qtbot.waitActive(widget, timeout=500):
+                show_action()
+
+        :param QWidget widget:
+            Widget to wait on.
```
The-Compiler: on -> for?

pytestqt/qtbot.py
```diff
@@ -141,6 +145,66 @@ def addWidget(self, widget):
...
+
+        :param int|None timeout:
+            How many miliseconds to wait on.
```
The-Compiler: miliseconds -> milliseconds; on -> for?

pytestqt/qtbot.py
```diff
@@ -141,6 +145,66 @@ def addWidget(self, widget):
...
+            ``PyQt4`` or ``PySide``.
+
+        .. note:: This method is also available as ``wait_active`` (pep-8 alias)
+        """
+        __tracebackhide__ = True
+        return _WaitWidgetContextManager('qWaitForWindowActive', 'activated', widget, timeout)
+
+    wait_active = waitActive  # pep-8 alias
+
+    def waitExposed(self, widget, timeout=1000):
```
The-Compiler: Same changes as above in this docstring :wink:

pytestqt/qtbot.py
```diff
@@ -150,17 +214,14 @@ def waitForWindowShown(self, widget):
         :param QWidget widget:
             Widget to wait on.
 
-        .. note:: In Qt5, the actual method called is qWaitForWindowExposed,
-            but this name is kept for backward compatibility
+        .. note:: In ``PyQt5`` this function is considered deprecated in favor of :meth:`waitForWindowExposed`.
```
The-Compiler: Should we mention `waitExposed` instead?

pytestqt/qtbot.py
```diff
@@ -495,8 +560,19 @@ def result(*args, **kwargs):
                 setattr(cls, method_name, method)
 
 
-# provide easy access to SignalTimeoutError to qtbot fixtures
+class TimeoutError(Exception):
```
The-Compiler: What about inheriting from `SignalTimeoutError` (for backwards compatibility) and deprecating `SignalTimeoutError`, like discussed in https://github.com/pytest-dev/pytest-qt/pull/157#issuecomment-241227483 some while ago?

pytestqt/qtbot.py
```diff
@@ -527,3 +603,40 @@ def _iter_widgets(item):
...
+        __tracebackhide__ = True
+        try:
+            if exc_type is None:
+                method = getattr(qt_api.QtTest.QTest, self._method_name)
+                r = method(self._widget, self._timeout)
+                if not r:
+                    msg = 'widget {} not {} in {} ms.'.format(self._widget, self._adjective_name, self._timeout)
+                    raise TimeoutError(msg)
+        finally:
+            self._widget = None
```
The-Compiler: Is this to avoid leaking memory?

pytestqt/qtbot.py
nicoddemus: Done, well remembered! 😁

nicoddemus:
I implemented all the changes you commented @The-Compiler!

The-Compiler:
Thanks! :+1:

PR #162, opened 2016-10-25
Use setuptools_scm for version management

nicoddemus:
Next step is automatic publishing to PyPI by pushing tags.


nicoddemus:
Pushed to pytest-dev/pytest-qt instead of my fork to make sure [readthedocs](http://pytest-qt.readthedocs.io/en/setuptools_scm/) built without errors.

PR #164, opened 2016-10-26
Run a window manager on travis

### LINKED ISSUE #160: Failure on travis: waitActive time-outs
`test_wait_window` for `waitActive` on the `py34-qt5` environment on Travis fails with:

```
=================================== FAILURES ===================================
______________________ test_wait_window[waitActive-True] _______________________
show = True, method_name = 'waitActive'
qtbot = <pytestqt.qtbot.QtBot object at 0x7ff758477780>
    @pytest.mark.parametrize('show', [True, False])
    @pytest.mark.parametrize('method_name', ['waitExposed', 'waitActive'])
    def test_wait_window(show, method_name, qtbot):
        """
        Using one of the wait-widget methods should not raise anything if the widget
        is properly displayed, otherwise should raise a TimeoutError.
        """
        method = getattr(qtbot, method_name)
        if qt_api.pytest_qt_api != 'pyqt5':
            with pytest.raises(RuntimeError) as exc_info:
                with method(None, None):
                    pass
            assert str(exc_info.value) == 'Available in PyQt5 only'
        else:
            widget = qt_api.QLineEdit()
            qtbot.add_widget(widget)
            if show:
                with method(widget, timeout=1000):
                    widget.move(100, 100)
                    widget.resize(100, 100)
                    widget.setFocus()
>                   widget.show()
E                   pytestqt.qtbot.TimeoutError: widget <PyQt5.QtWidgets.QLineEdit object at 0x7ff7584cb828> not activated in 1000 ms.
```

It was tried to move and resize the window based on the docs for `qWaitForWindowActive`:

```c++
    inline static bool qWaitForWindowActive(QWindow *window, int timeout = 5000)
    {
        QElapsedTimer timer;
        timer.start();
        while (!window->isActive()) {
            int remaining = timeout - int(timer.elapsed());
            if (remaining <= 0)
                break;
            QCoreApplication::processEvents(QEventLoop::AllEvents, remaining);
            QCoreApplication::sendPostedEvents(0, QEvent::DeferredDelete);
            QTest::qSleep(10);
        }
        // Try ensuring the platform window receives the real position.
        // (i.e. that window->pos() reflects reality)
        // isActive() ( == FocusIn in case of X) does not guarantee this. It seems some WMs randomly
        // send the final ConfigureNotify (the one with the non-bogus 0,0 position) after the FocusIn.
        // If we just let things go, every mapTo/FromGlobal call the tests perform directly after
        // qWaitForWindowShown() will generate bogus results.
        if (window->isActive()) {
            int waitNo = 0; // 0, 0 might be a valid position after all, so do not wait for ever
            while (window->position().isNull()) {
                if (waitNo++ > timeout / 10)
                    break;
                qWait(10);
            }
        }
        return window->isActive();
    }

``` 

But so far no good.

The-Compiler:
(Hopefully) fixes #160

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,8 @@ env:
 install:
  - sudo apt-get update
 
- # Xvfb
- - sudo apt-get install -y xvfb
+ # Xvfb / window manager
+ - sudo apt-get install -y xvfb herbstluftwm
 
  # Qt
  - python scripts/install-qt.py
@@ -38,10 +38,16 @@ install:
  # others
  - pip install coveralls --use-wheel
 
+before_script:
+ - "export DISPLAY=:99.0"
+ - "sh -e /etc/init.d/xvfb start"
+ - sleep 3 # give xvfb some time to start
+ - "herbstluftwm &"
+ - sleep 1
+
 script:
  - python setup.py develop
- # change screen size and depth in attempt to fix waitActive test (#160)
- - catchsegv xvfb-run --server-args="-screen 0 640x480x24" coverage run --source=pytestqt setup.py test
+ - catchsegv coverage run --source=pytestqt setup.py test
  - tox -e lint
 
 after_success:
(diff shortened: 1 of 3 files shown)


nicoddemus:
Ohhh yeah! 👍 

Many thanks! 😁

PR #165, opened 2016-10-26
Install PyQt5 from PyPI

The-Compiler:
Since PyQt5 5.6 it's available as binary wheel including Qt on PyPI.
This will make it easier for contributors too, as they can simply run
tox.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ dist: trusty
 python:
   - "2.7"
   - "3.4"
+  - "3.5"
 
 virtualenv:
   system_site_packages: true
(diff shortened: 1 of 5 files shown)


nicoddemus:
Thanks!

Apparently the pyqt5 binaries are not available for `py34`. If that's the case, I think it is OK to drop testing `py34` with `pyqt5`.

The-Compiler:
Alternatively I could use the distribution's PyQt5 for Python 3.4 - what would you prefer?

nicoddemus:
I'm OK with dropping py34-pyqt5 to simplify things. py34 and py35 are pretty much compatible, plus we are testing py34 with pyside.

To recap:

**Windows**

* py34-pyqt5 (FAILED)
* py35-pyqt5 (OK)
* py34-pyqt4 (OK)
* py27-pyqt4 (OK)
* py27-pyside (OK)
* py34-pyside (OK)

**Linux**

* py27-pyqt4 (OK)
* py27-pyside (OK)
* py34-pyside (OK)
* py27-pyside (OK)
* py34-pyqt5 (FAILED)

So I would be OK with dropping `py34-pyqt5` in both platforms and add `py35-pyqt5` to Travis instead. What do you think?

The-Compiler:
Sounds good, I hope I adjusted everything correctly! I also adjusted the text in the README.

nicoddemus:
No longer relevant now that we test with conda 😁

PR #167, opened 2017-02-08
Fix API error message and tests on Python 3.6

The-Compiler:
The error was wrong:

```
INTERNALERROR> RuntimeError: Invalid value for $PYTEST_QT_API: <pytestqt.qt_compat._QtApi object at 0x7f00b76f2198>
```

I also fixed some test breakage on Python 3.6 and added it to Travis - let me know if you want a separate PR for that instead :wink:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ dist: trusty
 python:
   - "2.7"
   - "3.4"
-  - "3.6"
 
 virtualenv:
   system_site_packages: true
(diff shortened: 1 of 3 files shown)


The-Compiler:
Looks like Travis doesn't have Python 3.6 with system-site-packages. Removed it for now. Probably can re-add it when I pick up #165 again.

nicoddemus:
Thanks Florian! All good to be in the same PR, don't worry.

I will review this ASAP, thanks again! 😉

nicoddemus:
Thanks @The-Compiler! 😁

PR #168, opened 2017-02-21
Fix issue where pytestqt was hiding the information when there's an exception raised from another exception on Python 3

fabioz:
This change makes a change to use print_exception instead of rolling your own when on Python 3. I've done the change on the fly here, so, it may need better tests on Python 3 (and alternatively, the same could be used on Python 2, but in that case io.StringIO may not work and may need StringIO.StringIO).

### DIFF
--- a/pytestqt/exceptions.py
+++ b/pytestqt/exceptions.py
@@ -70,8 +70,7 @@ def format_captured_exceptions(exceptions):
     message = 'Qt exceptions in virtual methods:\n'
     message += '_' * 80 + '\n'
     for (exc_type, value, tback) in exceptions:
-        message += ''.join(traceback.format_tb(tback)) + '\n'
-        message += '%s: %s\n' % (exc_type.__name__, value)
+        message += ''.join(traceback.format_exception(exc_type, value, tback)) + '\n'
         message += '_' * 80 + '\n'
     return message
 


pytestqt/exceptions.py
```diff
@@ -70,8 +70,16 @@ def format_captured_exceptions(exceptions):
     message = 'Qt exceptions in virtual methods:\n'
     message += '_' * 80 + '\n'
     for (exc_type, value, tback) in exceptions:
-        message += ''.join(traceback.format_tb(tback)) + '\n'
-        message += '%s: %s\n' % (exc_type.__name__, value)
+        if sys.version_info[0] > 2:
+            from io import StringIO
+            import traceback
+            s = StringIO()
+            traceback.print_exception(type_, value, tback, file=s)
```
The-Compiler: Why not `format_tb` as well?

fabioz: format_tb hides information on the cause of another exception (so, if an exception is raised from another exception, the cause is lost).

pytestqt/exceptions.py
```diff
@@ -67,11 +67,19 @@ def format_captured_exceptions(exceptions):
...
+    import traceback
     message = 'Qt exceptions in virtual methods:\n'
     message += '_' * 80 + '\n'
     for (exc_type, value, tback) in exceptions:
-        message += ''.join(traceback.format_tb(tback)) + '\n'
-        message += '%s: %s\n' % (exc_type.__name__, value)
+        if sys.version_info[0] > 2:
+            from io import StringIO
+            s = StringIO()
+            traceback.print_exception(type_, value, tback, file=s)
```
The-Compiler: Oh - I misread the code - why not `format_exception`?

The-Compiler:
You pushed while I was reviewing, but what I said is still valid :wink:

fabioz:
Agreed. format_exception seems better (and should work on both py2 and py3). Just submitted the change (sorry for the multiple commits -- I was too lazy to download it locally as I have just a conda installed version locally and just did it on the github ui... I think that on apply you can squash it all on a single commit).

nicoddemus:
Thanks @fabioz, squashed and rebased on `master`.

Anything else here @The-Compiler? Otherwise I think we can cut a `2.1.2` release.

The-Compiler:
I kinda forgot about this, but I think it's fine.

PR #174, opened 2017-07-11
Fix query of module version in docs

ghisvail:


ghisvail:
The problem with using `pkg_resources` is that one cannot build the documentation from a fresh checkout without generating the setuptools metadata first. Instead, since `setuptools_scm` will generate a version file in the final tarball, it is possible to do a plain import in `conf.py` (successfully tested on the Debian package).

nicoddemus:
Thanks! The failures on travis are unrelated, I will investigate them later.

PR #175, opened 2017-07-11
Fix master on Travis

nicoddemus:
Using system python directly, given that the old approach no longer works. 

This is related to the recently published new trusty image, given that setting `group: deprecated-2017Q2` made the build work again in my tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,16 @@
-language: python
+language: generic
 sudo: required
 dist: trusty
 
-python:
-  - "2.7"
-  - "3.4"
-
-virtualenv:
-  system_site_packages: true
-
-matrix:
-  exclude:
-    - env: PYTEST_QT_API=pyqt5
-      python: "2.7"
-
 env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -27,28 +19,28 @@ install:
...
  - sleep 3 # give xvfb some time to start
  - "herbstluftwm &"
  - sleep 1
 
 script:
- - python setup.py develop
- - catchsegv coverage run --source=pytestqt setup.py test
- - tox -e lint
+ - sudo catchsegv coverage run --source=pytestqt setup.py test
+ - sudo tox -e lint
```
The-Compiler: Why do those need `sudo`?

nicoddemus: `setuptools` (or `distutils`) would complain about not being able to write some files somewhere, I don't remember where exactly... TBH I didn't investigate it too much. 😅

The-Compiler:
Well, if it works, why not 😆

Issue #187, opened 2017-09-24
Properly clear exceptions after test


fabioz:
If an exception is thrown and is captured by pytestqt, it's keep alive (found it while trying to discover why PySide was crashing on the process exit when a test failed -- the reason was that pytestqt was keeping variables alive longer than it should and variables weren't cleared in the proper order because of that).

Will provide a fix shortly.

fabioz:
Is this fixed now?

nicoddemus:
Oh yeah, thanks for the ping. I have fixed the problem in pytest, we should be able to produce a proper test now.

nicoddemus:
Thanks @fabioz, and sorry for the delay!

fabioz:
No problems. Thanks for the fix ;)

PR #188, opened 2017-09-24
Don't keep exceptions alive after the test finishes

fabioz:
Clearing the exceptions found during a test is needed to allow variables to be properly cleared.

Fixes https://github.com/pytest-dev/pytest-qt/issues/187


nicoddemus:
Hi @fabioz, thanks for working on this!

While your fix seems correct, I think there's more to it. I was writing a test for your change (which I will push shortly) and notice that I couldn't make the test pass and discovered a bug in pytest itself, see pytest-dev/pytest#2798. After that is fixed we can probably make your PR pass.

PR #190, opened 2017-10-04
updated modeltester to allow QBrush as BackgroundColorRole

p0las:
pyQt5 takes QBrush as BackgroundColorRole and TextColorRole.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ dist: trusty
 env:
   global:
       # used by ci-helpers
-      - CONDA_CHANNELS=conda-forge SETUP_XVFB=true  DEPS="pytest tox coveralls"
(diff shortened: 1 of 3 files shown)


nicoddemus:
2.2.1 published! Thanks @p0las! 😁

The-Compiler:
I was in bed while all this happened, great work! 👍 

We should probably have a test for this though, no?

p0las:
@The-Compiler I have been thinking the same. Fix without a failing test first is not a great one.

nicoddemus:
Yep, completely missed that point for some reason. 😅 

@p0las would you like to work on this? I assume we just need a "model" which returns a `QBrush` for those roles.

nicoddemus:
Created: #191

PR #193, opened 2017-11-20
Add a qapp_args fixture

The-Compiler:
This makes it possible to pass custom arguments to Qt.

- I'm not sure how to test this yet - I can try something with `QCoreApplication.arguments()` and a subprocess, but I'm not sure whether it works.
- I'm also not sure where to document it, given that `qapp` isn't documented at all either. What do you think?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,10 @@
+2.3.0
+-----
+
+- New ``qapp_args`` fixture which can be used to pass custom arguments to
+  ``QApplication``.
+  Thanks `@The-Compiler`_ for the PR.
+
 2.2.1
 -----
 
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -3,6 +3,7 @@ Reference
 
 QtBot
 -----
+
 .. module:: pytestqt.qtbot
 .. autoclass:: QtBot
 
@@ -32,3 +33,10 @@ Record
 
 .. module:: pytestqt.logging
 .. autoclass:: Record
+
+qapp fixture
+------------
+
+.. module:: pytestqt.plugin
+.. autofunction:: qapp
+.. autofunction:: qapp_args
--- a/pytestqt/plugin.py
+++ b/pytestqt/plugin.py
@@ -21,7 +21,16 @@ assert format_captured_exceptions
 @pytest.fixture(scope='session')
 def qapp_args():
     """
-    Fixture providing QApplication arguments to use.
+    Fixture that provides QApplication arguments to use.
+
+    You can override this fixture to pass different arguments to
+    ``QApplication``:
+
+    .. code-block:: python
+
+       @pytest.fixture(scope='session')
+       def qapp_args():
+           return ['--arg']
     """
     return []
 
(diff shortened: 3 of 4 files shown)


nicoddemus:
Hey @The-Compiler!

> I can try something with QCoreApplication.arguments() and a subprocess, but I'm not sure whether it works.

I think this is the way to go. Perhaps printing `QCoreApplication.arguments()` and watching the output would be enough?

> I'm also not sure where to document it, given that qapp isn't documented at all either. What do you think?

I think adding a reference to https://github.com/pytest-dev/pytest-qt/blob/master/docs/reference.rst is enough (for both `qapp` and `qapp_args`).

The-Compiler:
Hmmm - things seem to work fine on Travis so far (and for me locally), but [fail]( https://ci.appveyor.com/project/nicoddemus/pytest-qt/build/1.0.458/job/4ocj5hbwu23u7h8q) on AppVeyor:

```
>       assert qapp.arguments() == ['--test-arg']
E       AssertionError: assert <PyQt4.QtCore...0000003F97DD8> == ['--test-arg']
```

Unfortunately the ellipsis don't really help here :laughing:. Do you know what's going on there? Do we get a `QStringList` or something?

nicoddemus:
> Unfortunately the ellipsis don't really help here 😆. Do you know what's going on there? Do we get a QStringList or something?

Weird. Well I changed the code to explicitly making a `list` and call `str()` on each argument, let's see how that goes.

The-Compiler:
I think there are two different issues here:

- The arguments are a string list in some scenarios. I can actually reproduce this when trying with PyQt by hand, and it looks like doing `list()` without calling `str()` on the arguments suffices:

```python
>>> from PyQt4.QtGui import QApplication
>>> app = QApplication(['--test-arg'])
>>> app.arguments()
<PyQt4.QtCore.QStringList object at 0x7f6dde415500>
>>> list(app.arguments())
[PyQt4.QtCore.QString(u'--test-arg')]
>>> '--test-arg' in list(app.arguments())
True
```

So I guess we can back out your last commit :wink:

- In some environments, we get the `sys.argv` passed to pytest instead of the arguments passed to Qt. This seems like a Windows + Qt4 thing? I wonder whether we should just skip the test there. I wonder how PySide2/PyQt5 look, but I'm off to bed now.

nicoddemus:
> I wonder whether we should just skip the test there

Sounds reasonable I guess.

The-Compiler:
Phew - after needing three tries to get a working version checks, things look fine now 😆 

The PySide2 errors are unrelated:

```
=================== 277 passed, 34 skipped in 28.79 seconds ====================
Fatal Python error: deallocating None
Current thread 0x00007fb949183740 (most recent call first):
Aborted (core dumped)
```

Working on a project which supports 3.5 Qt wrappers with 2 major Python versions on 2+ operating systems sure isn't easy sometimes 😆 

This is ready now as far as I'm concerned.

nicoddemus:
Thanks @The-Compiler! 

> Working on a project which supports 3.5 Qt wrappers with 2 major Python versions on 2+ operating systems sure isn't easy sometimes

Indeed! 😆 

> This is ready now as far as I'm concerned.

Yep! I will take a closer look at the errors when I get home, but either way we can probably merge this and push a release. Would you like to do the honors? All is needed is to push a tag and Travis will publish the package.

The-Compiler:
Sure, let's see how that goes! 🎉

The-Compiler:
Looks like [that worked](https://pypi.python.org/pypi/pytest-qt/2.3.0)! Do you usually write an announce mail or something?

nicoddemus:
I usually mention it on twitter, adding a link to the changelog. Want to do the honours here as well? 😉

The-Compiler:
Hah, looks you were faster! :wink:

nicoddemus:
I thought you were asleep by now so decided to send the tweet myself. 😝

PR #196, opened 2017-12-24
env variable wins over config value

mochick:
i believe that in the awkward case where you have `qt_api` and the `PYTEST_QT_API` set at the same time, the env variable should win according to the documentation
https://pytest-qt.readthedocs.io/en/latest/intro.html


The-Compiler:
Don't worry about the failing PySide2 tests, those are unrelated to your change.

However, can you please also add a test for this scenario where both values are set?

nicoddemus:
Thanks @mochick and sorry for the delay!

Issue #170, opened 2017-05-11
InvocationError when testing via tox but not via pytest called directly


ixjlyons:
I've set up a minimally reproducible example as a repo [here](https://github.com/ixjlyons/qtbottest). The README explains how I've invoked the test using pytest directly and with tox, and I can't figure out why it doesn't work with tox.

If I've made an obvious error in my tox.ini somewhere, I'd appreciate finding out what to fix. [This](https://github.com/tox-dev/tox/blob/2b2a51f8cd845fb6df3f87bb6e83f3e9cc72ab28/tox/session.py#L199) is the only place in the tox code that prints out the `InvocationError` without any other information except the command in question (what I'm seeing).

It should also be noted that I can run pytest-qt's tests with tox just fine. Also, if I comment out the test function's contents and remove the use of the `qtbot` fixture, the test runs fine with tox.

The-Compiler:
Does adding `passenv = *` under `[testenv]` change anything? Can you install the `pytest-faulthandler` plugin and run pytest with `-s` and see if that gives you anything useful?

ixjlyons:
`passenv = DISPLAY XAUTHORITY` seems to do the trick (both required). Should a note about using pytest-qt with tox (I'm assuming also it's Linux-specific) be added to the docs?

@The-Compiler thanks a lot for the quick help by the way. This issue has irritated me for months and I only now put in a bit of time to produce a demo to try and get it resolved.

The-Compiler:
You'll either need to do that or install [pytest-xvfb](https://github.com/The-Compiler/pytest-xvfb) then.

Unfortunately, Qt simply calls `abort()` on that error, and pytest's output capturing can't show the error to you as the process was just killed.

There's really not much we can do about this... I wonder if it'd help if pytest-qt disabled the output capturing while creating the `QApplication` though.

nicoddemus:
@ixjlyons perhaps we can add a "Troubleshooting" sections to the docs, where we can mention both the "tox passenv" solution and the "abort()" issue?

cauveryguda:
I am facing the same issue with tox.  Tried placing passenv = DISPLAY XAUTHORITY  under testenv in tox.ini but that did not help. Did your help on fixing this.

The-Compiler:
@cauveryguda Is that with pytest? Can you run pytest with `-s` and see what's printed?

reinjunon:
[...]

PR #198, opened 2018-01-17
Troubleshooting section with InvocationError/missing DISPLAY

ederag:
As suggested in https://github.com/pytest-dev/pytest-qt/issues/170#issuecomment-301060437, here is a tentative troubleshooting section. Thanks to [ixjlyons](https://github.com/ixjlyons) for tracking that down, and to [The-Compiler](https://github.com/The-Compiler) for the solution !

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,6 +25,7 @@ pytest-qt
     modeltester
     app_exit
     note_pyqt4v2
+    troubleshooting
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks a lot @ederag, we appreciate you taking the time to tackle this! 😁

Issue #19, opened 2014-09-16
Add an example on how to handle modal dialogs during testing


nicoddemus:
As discussed in #18

nicoddemus:
Fixed by #201

Issue #18, opened 2014-09-15
behaviour with `QMessageBox.question` and other modal dialogs


baudren:
I do not understand how to successfully test the behaviour of modal dialogs, and did not find an example in the documentation -- or anywhere else, for that matter.
## Question

How to prevent modal dialogs from poping up in testing, and how do I simulate clicking on their buttons with `qtbot`?
## Sample file

``` python
from PySide.QtGui import QMessageBox, QFrame

class Simple(QFrame):
    def query(self):
        self.question = QMessageBox.question(
            self, 'Message', 'Are you sure?',
            QMessageBox.Yes | QMessageBox.No,
            QMessageBox.No)
        if self.question == QMessageBox.Yes:
            self.answer = True
        else:
            self.answer = False

def test_Qt(qtbot):
    simple = Simple()
    qtbot.addWidget(simple)

    simple.query()
```
## What I tried

Because `self.question` is a QMessageBox.StandardButton, _i.e._ an enum, it has no method to allow clicking on it. I tried to set directly `self.question` to `QMessageBox.Yes`, for instance, but obviously, this does not close the window.

Also, adding `self.question` to `qtbot` with `addWidget` does not work, as it is no real widget.

Could this be due to a wrong design choice on my side? Should the opening of the question be in a separate function, and the rest of the logic would simply take as an input the result of the `question`. Or am I testing this the wrong way?
## Additional but (maybe) related issues

It is seems related to me, but when calling an `exec_` method on a customized `QtGui.QDialog` object, the same problem appears. I can not use the method `qtbot.mouseClick` on the object before clicking on it manually. Because the object pops out, and interrups the flow of the program.

Of course, this is the whole point of such an object in the main application, but while testing, it looks like it messes things up. Feel free to ignore this part if you deem it too unrelated (I can also open another ticket).

PR #201, opened 2018-03-13
Add a note about how to handle modal dialogs.

HazenBabcock:
This I think adds the modal dialog handling documentation you were looking for in issue #18 and issue #19.

### DIFF
--- a/docs/note_dialogs.rst
+++ b/docs/note_dialogs.rst
@@ -56,5 +56,3 @@ And now it is also easy to mock ``AskNameAndAgeDialog.ask`` when testing the for
         # calls AskNameAndAgeDialog.ask
         # test that the rest of the form correctly behaves as if
         # user entered "Jonh" and 30 as name and age
-
-These examples use the ``pytest-mock`` package.
--- a/docs/note_dialogs.rst
+++ b/docs/note_dialogs.rst
@@ -4,15 +4,15 @@ A note about Modal Dialogs
 Simple Dialogs
 --------------
 
-For QMessageBox.question one approach is to ``mock`` it:
+For QMessageBox.question one approach is to mock the function using the ``monkeypatch`` fixture:
 
 .. code-block:: python
                 
-    def test_Qt(qtbot, mock):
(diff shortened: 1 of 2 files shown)


docs/note_dialogs.rst
```diff
@@ -0,0 +1,60 @@
...
+==========================
+
+Simple Dialogs
+--------------
+
+For QMessageBox.question one approach is to ``mock`` it:
+
+.. code-block:: python
+                
+    def test_Qt(qtbot, mock):
```
The-Compiler: This should be called `mocker`, `mock` is just a compatibility alias: https://github.com/pytest-dev/pytest-mock/issues/4

However, why require another plugin just for a doc example anyways? I think it'd be better to just use the built-in [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html).

docs/note_dialogs.rst
```diff
@@ -0,0 +1,60 @@
...
+
+   name, age = AskNameAndAgeDialog.ask("Enter name and age because of bananas:", parent)
+   if name is not None:
+       # use name and age for bananas
+
+And now it is also easy to mock AskNameAndAgeDialog.ask when testing the form:
+
+.. code-block:: python
+                
+    def test_form_registration(qtbot, mock):
```
The-Compiler: Same as above regarding `mock`.

nicoddemus:
Thanks a lot @HazenBabcock!

I agree with @The-Compiler's comments so I went ahead and applied the changes myself.

Thanks again for submitting the PR, we appreciate it. 👍

nicoddemus:
Fixed #19

HazenBabcock:
Great, thanks for the help with this PR.

nicoddemus:
@The-Compiler are you OK with merging this? 😁 (failures are unrelated and I plan to tackle them in #202)

The-Compiler:
Pushed a commit to remove the last sentence as it's not true anymore. LGTM now! :+1:

nicoddemus:
Thanks!

Issue #202, opened 2018-03-15
PySide2 segmentation faults after test session in Travis and AppVeyor


nicoddemus:
Sample builds:

* https://ci.appveyor.com/project/nicoddemus/pytest-qt/build/1.0.480
* https://travis-ci.org/pytest-dev/pytest-qt/builds/353613828

nicoddemus:
The jobs are marked as "allowed to fail" on Travis and removed from AppVeyor for now.

rth:
Possibly due to https://bugreports.qt.io/browse/PYSIDE-585 (fixed in the dev version of PySide2).

nicoddemus:
@rth thanks for the heads up, appreciate it!

The-Compiler:
This doesn't seem to be an issue nowadays anymore.

PR #203, opened 2018-03-15
Allow failures of PySide 2 builds for now

nicoddemus:
At least until we can investigate #202 or it solves itself in new
versions of pyside2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
 
 matrix:
   # PySide2 crashes: #202
-  allow-failures:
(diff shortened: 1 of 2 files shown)


The-Compiler:
This should be `allow_failures` and not `allow-failures` on Travis. Let's see if it works now...

nicoddemus:
Ahh thanks! I blame the fact that it was late and I was about to go to bet. 😁

nicoddemus:
(Merging given that AppVeyor had passed already)

PR #207, opened 2018-04-27
DOC: Add troubleshoot entry for UI Events when using xvfb and Travis-CI.

hhslepicka:


nicoddemus:
Awesome, thank you very much, we really appreciate it!

Issue #205, opened 2018-04-13
QStringModel not in QtGui but QtCore on Pyside2


skoudoro:
Hi, 

When I run `pytest`, I got this error:

`INTERNALERROR> AttributeError: module 'PySide2.QtGui' has no attribute 'QStringListModel'`

The doc [here](https://doc-snapshots.qt.io/qtforpython/PySide2/QtGui/QStringListModel.html) is a bit confusing for `QStringListModel` but this model seems to be in QtCore.

Furthermore, it seems this model is not present in all pyside2 as you can see [here](https://wiki.qt.io/PySide2_Missing_Bindings)

am I right or did I do something wrong?

nicoddemus:
Hi @skoudoro, thanks for writing.

Can you please post the full traceback, along with python and PySide2 versions you are using?

For the record our CI uses PySide2 version 2.*:

https://github.com/pytest-dev/pytest-qt/blob/dcfc031aa248230fd22026881ea5a013d74f1c6b/.travis.yml#L23

And our latest build on Travis used `2.0.0.dev0`:

https://travis-ci.org/pytest-dev/pytest-qt/jobs/353785654

skoudoro:
ok, Thanks for this information @nicoddemus. Here requested infos : 
windows 10 - python==3.6.4 - pysside2==5.9.0a1 - pytest-qt from conda-forge
I install pyside2 via `pip install --index-url=http://download.qt.io/snapshots/ci/pyside/5.9/latest/ pyside2 --trusted-host download.qt.io`

TraceBack:
-----------
```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\mpsts-env-3\lib\site-packages\_pytest\main.py", line 103, in wrap_session
INTERNALERROR>     config._do_configure()
INTERNALERROR>   File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\mpsts-env-3\lib\site-packages\_pytest\config.py", line 914, in _do_configure
INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
INTERNALERROR>   File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\mpsts-env-3\lib\site-packages\pluggy\__init__.py", line 628, in call_historic
[...]

PR #209, opened 2018-05-15
Fix QStringListModel error on PySide2

rth:
Fixes https://github.com/pytest-dev/pytest-qt/issues/205

On PySide2,  `QStringModel` is not in QtGui but in QtCore, this applies the fix suggested by @nicoddemus

### DIFF
--- a/pytestqt/qt_compat.py
+++ b/pytestqt/qt_compat.py
@@ -108,13 +108,15 @@ class _QtApi:
             self.Signal = QtCore.Signal
             self.Slot = QtCore.Slot
             self.Property = QtCore.Property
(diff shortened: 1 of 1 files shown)


rth:
> Coverage decreased (-0.08%)

We testing this change would require for the CI to fail without it with PySide2. Currently, it doesn't presumably because PySide2 wheels are more up to date that the PySide2 on conda so unless we add another Travis CI job that installs PySide2 with pip there is not much I can do about this ...

nicoddemus:
No worries @rth, thanks!

nicoddemus:
Thanks for the PR, appreciate it!

PR #212, opened 2018-05-21
Use Item.get_closest_marker on pytest 3.6

The-Compiler:
This avoids deprecation warnings on the upcoming pytest 3.6 release:
https://docs.pytest.org/en/features/mark.html#marker-revamp-and-iteration


The-Compiler:
pytest 3.6.0 [is out now](https://pypi.org/project/pytest/#history) - so I'd appreciate a release :smile:

Not sure if I could even do that myself (by pushing a tag?) - but if you feel like it, please go ahead!

PR #216, opened 2018-05-30
Properly handle chained exceptions in Python 3

### LINKED ISSUE #215: pytest-qt is not printing the cause of exceptions on python 3
Consider the code below: The line `Error 1` should appear in the stderr, but only `Error 2` appears.

This happens because `exceptions.format_captured_exceptions` is a custom exception printer that doesn't take into account the exception cause.

```
from pytestqt.qt_compat import qt_api

class MyWidget(qt_api.QWidget):

    def method(self):
        raise RuntimeError('Error 1')

    def event(self, ev):
        called.append(1)
        try:
            self.method()
        finally:
            raise RuntimeError('Error 2')


weak_ref = None
called = []


def test_1(qapp):
    global weak_ref
    w = MyWidget()
    qapp.postEvent(w, qt_api.QEvent(qt_api.QEvent.User))
    qapp.processEvents()
    assert len(called) > 0
    w.deleteLater()
```

nicoddemus:
Fix #215

### DIFF
--- a/pytestqt/exceptions.py
+++ b/pytestqt/exceptions.py
@@ -68,12 +68,19 @@ def format_captured_exceptions(exceptions):
     Formats exceptions given as (type, value, traceback) into a string
     suitable to display as a test failure.
     """
-    message = 'Qt exceptions in virtual methods:\n'
-    message += '_' * 80 + '\n'
+    if sys.version_info.major == 2:
+        from StringIO import StringIO
+    else:
+        from io import StringIO
+    
+    stream = StringIO()
+    stream.write('Qt exceptions in virtual methods:\n')
+    sep = '_' * 80 + '\n'
+    stream.write(sep)
     for (exc_type, value, tback) in exceptions:
-        message += ''.join(traceback.format_exception(exc_type, value, tback)) + '\n'
-        message += '_' * 80 + '\n'
-    return message
+        traceback.print_exception(exc_type, value, tback, file=stream)
+        stream.write(sep)
+    return stream.getvalue()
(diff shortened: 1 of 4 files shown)


pytestqt/exceptions.py
```diff
@@ -68,12 +68,19 @@ def format_captured_exceptions(exceptions):
     Formats exceptions given as (type, value, traceback) into a string
     suitable to display as a test failure.
     """
-    message = 'Qt exceptions in virtual methods:\n'
-    message += '_' * 80 + '\n'
+    if sys.version_info.major == 2:
+        from StringIO import StringIO
+    else:
+        from io import StringIO
```
The-Compiler: Why not have this import at the top of the file? Might also want to do `except ImportError:` instead so you don't need `sys` if you do it at the top.

nicoddemus: > Why not have this import at the top of the file?

I put it here locally mostly because this is function is rarely used. If you prefer I can move it to the top though, no problem.

> Might also want to do except ImportError: instead so you don't need sys if you do it at the top.

In general I tend to avoid that because it might lead to hard to diagnose erors (see https://github.com/pytest-dev/pytest-mock/issues/68).

nicoddemus: @The-Compiler gentle ping. 😁

The-Compiler: Fair point - seems fine to keep it here.

nicoddemus: Thanks!

PR #219, opened 2018-07-12
Remove mouseEvent proxy method

mitya57:
`QTest::mouseEvent` was never documented in Qt, and it is marked as internal in Qt source code:

https://github.com/qt/qtbase/blob/74305ba470f48da8b4c4e806fc714fe9f7649156/src/testlib/qtestcase.qdoc#L1379-L1387

It was also removed in PyQt v5.11, so `test_expected_qtest_proxies` fails with that release.


nicoddemus:
Thanks, appreciate it!

PR #234, opened 2018-09-11
Show import error messages for Qt backends

The-Compiler:
This makes it easier to debug why importing failed.

I tried to write a test for this, but gave up because patching `__import__` causes funny issues in pytest itself when importing things late there...

```diff
diff --git a/tests/test_basics.py b/tests/test_basics.py
index 847dfde..7f03645 100644
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -1,3 +1,4 @@
+import sys
 import weakref
 
 import pytest
@@ -459,3 +460,25 @@ def test_qapp_args(testdir):
     )
     result = testdir.runpytest_subprocess()
     result.stdout.fnmatch_lines(["*= 1 passed in *"])
+
+
+def test_importerror(monkeypatch):
+    real_import = __import__
+
+    def _fake_import(name, *args):
+        if name not in ['PySide', 'PySide2', 'PyQt4', 'PyQt5']:
+            return real_import(name, *args)
+        import pdb; pdb.set_trace()
+        raise ImportError("Failed to import {}".format(name))
+
+    module = '__builtins__' if sys.version_info.major == 2 else 'builtins'
+    monkeypatch.setattr('{}.__import__'.format(module), _fake_import)
+
+    expected = ("pytest-qt requires either PySide, PySide2, PyQt4 or PyQt5 to be installed\n"
+                "  PyQt4.QtCore: Failed to import PyQt4\n"
+                "  PyQt5.QtCore: Failed to import PyQt5\n"
+                "  PySide.QtCore: Failed to import PySide\n"
+                "  PySide2.QtCore: No module named PySide2")
+
+    with pytest.raises(RuntimeError, match=expected):
+        qt_api.set_qt_api(api=None)
```

Here's how the output looks now:

```
INTERNALERROR> RuntimeError: pytest-qt requires either PySide, PySide2, PyQt4 or PyQt5 to be installed
INTERNALERROR>   PyQt4.QtCore: No module named 'PyQt4'
INTERNALERROR>   PyQt5.QtCore: No module named 'PyQt5'
INTERNALERROR>   PySide.QtCore: No module named 'PySide'
INTERNALERROR>   PySide2.QtCore: No module named 'PySide2'
```

### DIFF
--- a/pytestqt/qt_compat.py
+++ b/pytestqt/qt_compat.py
@@ -70,9 +70,14 @@ class _QtApi:
     def set_qt_api(self, api):
         self.pytest_qt_api = self._get_qt_api_from_env() or api or self._guess_qt_api()
         if not self.pytest_qt_api:  # pragma: no cover
-            errors = '\n'.join('  {}: {}'.format(module, reason)
-                               for module, reason in sorted(self._import_errors.items()))
-            msg = "pytest-qt requires either PySide, PySide2, PyQt4 or PyQt5 to be installed\n" + errors
+            errors = "\n".join(
+                "  {}: {}".format(module, reason)
+                for module, reason in sorted(self._import_errors.items())
+            )
+            msg = (
+                "pytest-qt requires either PySide, PySide2, PyQt4 or PyQt5 to be installed\n"
+                + errors
+            )
             raise RuntimeError(msg)
 
         _root_modules = {
--- a/pytestqt/qt_compat.py
+++ b/pytestqt/qt_compat.py
@@ -24,6 +24,9 @@ class _QtApi:
     gets called, providing a uniform way to access the Qt classes.
     """
 
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks @The-Compiler!

About the test, I think we can have a `_import` function in `qt_compat.py` which just calls the builtin import:

```python
def _import(name):
    return __import__(name)
``` 

And use that indirection in `qt_compat.py`. Then, you can instead monkeypatch `qt_compat._import` to raise the errors you want. This way you don't need to mess with the builtin `__import__` at all. What do you think?

The-Compiler:
Great idea, didn't think of that! I added tests now. [Like they say](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering), *"All problems in computer science can be solved by another level of indirection"* :wink:

PR #237, opened 2018-09-13
Attempt to fix Travis 

nicoddemus:
Manually configure miniconda on Travis and use pytest-xvfb

Our travis build started failing with errors like this for pyqt 4 and 5:

```
UnsatisfiableError: The following specifications were found to be in conflict:
  - pyqt
  - pyqt=4.*
```
  
This commit installs miniconda manually so we can have better control at
what is going on, and install pytest-xvfb to handle xvfb for us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
       # used by ci-helpers
       - DEPS="pytest tox coveralls six"
       - MINICONDA_VERSION=latest
+      - DISPLAY=":99.0"
 
   matrix:
       - LINTING=1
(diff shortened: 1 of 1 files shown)


The-Compiler:
Yeah, see https://github.com/pytest-dev/pytest-qt/pull/124 and https://github.com/The-Compiler/pytest-xvfb/issues/1 - still no clue why it happens, but it's only on Qt 4. :wink:

nicoddemus:
Thanks for the pointer. I've read the related issues and indeed seems like our exact problem, although now we have one test failure in pyqt5 that we didn't have so far. I've reverted to running `xvfb` manually now, we revisit this later. 👍

nicoddemus:
The build in my fork has completed successfully: https://travis-ci.org/nicoddemus/pytest-qt/builds/428174533 🎉

nicoddemus:
Merging as this does not affect AppVeyor. 👍

Issue #221, opened 2018-07-13
Update model tester based on Qt changes


The-Compiler:
It looks like the Qt modeltester had [some changes](https://github.com/qt/qtbase/commits/5.11/src/testlib/qabstractitemmodeltester.cpp) (and a [bigger refactoring](https://github.com/qt/qtbase/commit/b3e4be2d8b9debf217657436139da0152f6f8797#diff-227bd803a8cca931ce30cdfbe060dd0e)) recently. We should probably try to adopt those.

This also means links to the old `modeltester.cpp` are invalid and need updating.

nicoddemus:
Thanks for the heads up @The-Compiler!

The-Compiler:
Opened a PR for this, but noticed after those changes that at least PyQt5 actually exposes the new QAbstractItemModelTester from Qt!

First prototype:

```diff
diff --git a/pytestqt/logging.py b/pytestqt/logging.py
index 0e8989c..c99ee35 100644
--- a/pytestqt/logging.py
+++ b/pytestqt/logging.py
@@ -182,6 +182,8 @@ class _QtMessageCapture(object):
         Method to be installed using qInstallMessageHandler (Qt5),
         stores each message into the `_records` attribute.
         """
+        if context.category == 'qt.modeltest':
+            pytest.fail(message)
         self._append_new_record(msg_type, message, context=context)
 
     @property
diff --git a/pytestqt/modeltest.py b/pytestqt/modeltest.py
index 2d7a095..f5dcb0b 100644
--- a/pytestqt/modeltest.py
+++ b/pytestqt/modeltest.py
@@ -87,6 +87,17 @@ class ModelTester:
         Whenever anything happens recheck everything.
         """
         assert model is not None
+        try:
+            tester = qt_api.QtTest.QAbstractItemModelTester(
+                model,
+                qt_api.QtTest.QAbstractItemModelTester.FailureReportingMode.Warning,
+            )
+        except AttributeError:
+            self._debug("Using Python tester")
+        else:
+            self._debug("Using Qt tester")
+            return
+
         self._model = model
         self._fetching_more = False
         self._insert = []
```

The problem is that we can't raise an `AssertionError` on failures, so most stuff in `test_modeltester.py` fails. Didn't have a good idea to make it pass in both scenarios yet, and not sure if I want to follow this further. With that, the output also looks suboptimal at the moment:

```
[...]

PR #238, opened 2018-09-13
Update modeltester to latest Qt sources

The-Compiler:
See #221

Last one for now :wink:

### DIFF
--- a/docs/modeltester.rst
+++ b/docs/modeltester.rst
@@ -36,15 +36,6 @@ items and call ``qtmodeltester.check``:
 If the tester finds a problem the test will fail with an assert pinpointing
 the issue.
 
-The following attribute may influence the outcome of the check depending
-on your model implementation:
-
-* ``data_display_may_return_none`` (default: ``False``): While you can
-  technically return ``None`` (or an invalid ``QVariant``) from ``data()``
-  for ``QtCore.Qt.DisplayRole``, this usually is a sign of
-  a bug in your implementation. Set this variable to ``True`` if this really
-  is OK in your model.
-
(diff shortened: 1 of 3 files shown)


The-Compiler:
I think I'll need some help with Qt 4 here. Looking at it, I don't think the type checks ever worked properly there, but our tests only tested wrong types and the `DisplayRole`, which happened to be added to the checked roles with the upstream Qt changes.

I'm not really sure what the right way to make this work is. I think before 41b83acae0db37eb1b073450f917e6db26154a1d the checks didn't make much sense at all, and c2403a02d38edd9664448fae07f3337d34d51836 seems to be needed for Python 2. But now I also see `QString` in the tests, so apparently the allowed types should be one of:

- `str` (Python 3)
- `unicode` (Python 2)
- `QString` (if available)

or something?

nicoddemus:
You are right that it seems previously nothing was tested properly.  

PyQt4 originally (the `pyqt4` API) always returned native `QString` objects. The V2 API (`pyqt4v2`), introduced later, returns native Python types instead. So in Python 2 I assume those are always `str` objects in some who-knows-what encoding. At work we do something like `qstring.utf8().decode('utf-8')` to obtain a `unicode` object.

Well it is quite a pickle. I pushed some tweaks to that in de8702b, let's see how it fares.

The-Compiler:
Oh, seems to look better now, thanks! Not sure if I should also add some more tests, but on the other hand I think the actual test coverage (i.e. "cases" of broken models we test) didn't really change much.

nicoddemus:
I guess we can add the changelog entry later before the release. 👍

PR #239, opened 2018-09-15
Use C++ modeltester if available

The-Compiler:
Based on #238, will rebase once that's merged. I'll let you be the judge re maintainance, it seems quite okay to me, and I'd really prefer the more thorough C++ tester if available.


nicoddemus:
I think this can now be rebased! 👍

The-Compiler:
Rebased to get the tests to run, probably will rebase again and fix up the changelog once #236 is in.

The-Compiler:
Hmm, 989f001dbcd24869afcc53770a69a1030d909bca seems to break because `rec.context` is None: https://travis-ci.com/The-Compiler/pytest-qt/jobs/147350627#L1295

I thought the previous code already did assume that `rec.context` isn't None?

nicoddemus:
You mean `context_format`? Yeah seems like it...

nicoddemus:
Oh it assumed `rec.context` was not `None` for Qt5 only (which indeed passes).

nicoddemus:
I think now that's left to do is the release. 😁

PR #240, opened 2018-09-15
Add wait argument to qtbot.assertNotEmitted

The-Compiler:
I swear this is the last one :see_no_evil: Take your time to review things though, I'm not in a hurry!

Needed this for some tests in qutebrowser recently, and it seems simple enough to add it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,9 @@
 
 - Improve debugging output when no Qt wrapper was found.
 - Register the ``no_qt_log`` marker with pytest so ``--strict`` can be used.
+- ``qtbot.assertNotEmitted`` now has a new ``wait`` parameter which can be used
+  to make sure asynchronous signals aren't emitted by waiting after the code in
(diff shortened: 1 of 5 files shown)


The-Compiler:
Whoops, accidentally had a `;` at the end of the line in the docs :laughing:  That's what I get for writing JavaScript, C++ and Python on the same day!

The-Compiler:
Thinking about this some more, I noticed a QEventLoop is spin up every time `assertNotEmitted` is used now (with the default `timeout=0`), which seems weird.

I've opted to just not spinning up an event loop with `qtbot.waitSignal(..., timeout=0)` at all, see the last commit - what do you think?

nicoddemus:
> I've opted to just not spinning up an event loop with qtbot.waitSignal(..., timeout=0) at all, see the last commit - what do you think?

Seems reasonable, thanks!

nicoddemus:
(and sorry for the delay!)

PR #241, opened 2018-09-23
Release v3.1.0

The-Compiler:


The-Compiler:
Tag pushed: https://travis-ci.org/pytest-dev/pytest-qt/builds/432208636

The-Compiler:
Huh, looks like Travis tries to release for two jobs and fails in one of them? Anyways, seems like that worked! :tada:

PR #242, opened 2018-09-26
Add CallbackBlocker.assert_called_with()

The-Compiler:
When changing my testsuite to use `CallbackBlocker`, I noticed checking args and kwargs individually is as list/dict is a bit cumbersome.

The API is inspired by `unittest.mock`, but I don't think we need the other stuff (like `assert_called_once_with`).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+3.2.0 (unreleased)
+------------------
+
+- The ``CallbackBlocker`` returned by ``qtbot.waitCallback()`` now has a new
+  ``assert_called_with(...)`` convenience method.
+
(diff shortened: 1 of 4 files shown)


nicoddemus:
Feel free to make a new release if you want to use this right away. 👍

The-Compiler:
Given that there's no changes to be done before the release, can I just push the tag without a PR?

nicoddemus:
We need to update the CHANGELOG with today's date: `3.2.0 (unreleased)` 😉

nicoddemus:
In the future if you want to speed things up, you can push the feature branch with the changelog already complete; once approved we can push the tag directly to the PR and release it in one go.

PR #245, opened 2018-10-01
Fix compatibility with PyQt5 5.11.3

The-Compiler:
Because of that assertion, we just break hard with PyQt5 5.11.3 (which is going to be released tomorrow or Wednesday) for any testsuite having pytest-qt installed.

IMHO, in the future we should prefer having a test which tests for the missing attribute instead of adding an assertion which breaks things for everyone :wink: 

I plan to push a tag directly after this is merged.


nicoddemus:
Oh definitely, brain fart on my part. 😝 

Please feel free to merge and push the tag (to the commit in this PR, not to the merge commit) when AppVeyor finishes. 👍

The-Compiler:
Done!

Issue #11, opened 2014-05-29
Catch exceptions in other Qt threads


jdreaver:
I recently switched to pytest-qt, and I love it so far. Before, I used my own pytest fixture to create a QApplication. I found a way to catch exceptions on all threads by overriding sys.excepthook in the fixture and checking for errors:

``` python
caught_exceptions = []


@pytest.yield_fixture
def qtbot_(request, qtbot):
    # Set excepthook to catch pyside errors in other threads.
    global caught_exceptions
    caught_exceptions = []

    def pytest_excepthook(type_, value, tback):
        global caught_exceptions
        caught_exceptions.append(''.join(traceback.format_tb(tback)) + "\n" +
                                 str(type_.__name__) + ": " + str(value))
        sys.__excepthook__(type_, value, tback)
    sys.excepthook = pytest_excepthook

    yield qtbot

    sys.excepthook = sys.__excepthook__
    if caught_exceptions:
        pytest.fail("Caught qt exceptions:\n{}".format(
            "\n".join(caught_exceptions)))
```

I created this so I can see when my table model methods fail. Without this, if an error happens in the `data()` or `header()` methods of a table model, like when a table view is being populated using the model, the test will not fail.

I think something similar to this would be a nice feature of pytest-qt, but a better, less hacky implementation would probably be better.

patrickkidd:
FWIW, it is probably more accurate to report these as "Exceptions caught in Qt event loop" rather than "Qt exceptions in virtual methods." This is because these exceptions can also be raised in a pyqt slot.

nicoddemus:
@patrickkidd good point, thanks. Implemented it in #255.

PR #255, opened 2019-02-27
Improve "uncaught exception" message by mentioning Qt event loop instead of virtual methods

nicoddemus:
As suggested in:

https://github.com/pytest-dev/pytest-qt/issues/11#issuecomment-467867594

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,11 @@
+3.2.3 (unreleased)
+------------------
+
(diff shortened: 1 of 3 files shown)


nicoddemus:
@blueyed have you used coveralls before? Any thoughts on how it compares with `codecov`?

blueyed:
@nicoddemus 
Pytest also has used it before IIRC.. :)
I prefer codecov (if it works).

nicoddemus:
👍

PR #279, opened 2019-12-07
Change to 'src' layout and drop Python 2

nicoddemus:


The-Compiler:
Yeaaaah! Finally away with Qt4 stuff :laughing:

Issue #224, opened 2018-08-24
Add docs on how to create custom QApplications


nicoddemus:
Follow up from https://github.com/pytest-dev/pytest-qt/issues/172#issuecomment-415588753

josiahlaivins:
I added the referenced code, and the py test starts up the application, but none of my qbot tests run.
```python
import sys
from typing import Union
import pytest
from ..app.pyside_app.components.Application import Application

# root conftest.py
this = sys.modules[__name__]
this.qapplication: Union[Application, None] = None

@pytest.fixture(scope='session')
def qapp():
	if this.qapplication is None: this.qapplication = Application(False)
	yield this.qapplication
```
Then I have an actual test:
```python
from pytestqt import qtbot

def test_iot_controller(qapp, qtbot: qtbot):
	print('hello')
	qapp.close()  # Where Application has a method:     def close(self):  self.exit()
```
But that test never gets executed. I also don't understand how to close the application.

nicoddemus:
> But that test never gets executed

That's strange, what's the name of the test file? What output are you getting?

> I also don't understand how to close the application.

Fixture teardown is done after the `yield`:

```python
@pytest.fixture(scope='session')
def qapp():
	if this.qapplication is None: this.qapplication = Application(False)
	yield this.qapplication
    # close this.qapplication here
```

m3nu:
Took a shot at adding this, by merging it with the existing `QApplication.exit()` docs. Let me know if you want any changes, @nicoddemus

PR #288, opened 2020-02-29
Clarify using QApplication in tests

m3nu:
This implements #224 by extending the existing documentation on `QApplication.exit()`


docs/qapplication.rst
```diff
@@ -0,0 +1,82 @@
...
+            pass
+
+
+If your tests require access to app-level functions, like
+``CustomQApplication.custom_function()``, you can override the built-in
+``qapp`` fixture to use your own app:
+
+.. code-block:: python
+
+    # in root conftest.py
```
nicoddemus: I think the whole code is better done like this:

```python
    @pytest.fixture(scope="session")
    def qapp():
        yield CustomQApplication([])
```

Being session-scoped means that pytest will take care of calling this exactly once.

docs/qapplication.rst
```diff
@@ -0,0 +1,82 @@
...
+            button = get_app_exit_button()
+            qtbot.click(button)
+            assert QApplication.exit.call_count == 1
+
+
+Testing Custom QApplications
+----------------------------
+
+It's possible to test custom ``QApplication`` classes, but you need to be
+careful to avoid multiple app instances in the same test. Assuming to defined
```
nicoddemus:
```suggestion
careful to avoid multiple app instances in the same test. Assuming one defines a
```

Assuming that's the original intent, of course.

docs/qapplication.rst
m3nu: Right. Simplified it as suggested. Still works OK with the other code removed.

nicoddemus:
Thanks a lot @m3nu, really appreciate this. 👍

m3nu:
Welcome. Thanks for providing this library and hoping to contribute more in the future, when I notice something.

PR #294, opened 2020-04-08
Enforce addWidget-argument to be a QWidget

### LINKED ISSUE #290: Add type checking for qtbot.addWidget()?
In a training I was giving, someone accidentally passed a model rather than the corresponding view into `qtbot.addWidget()`. When doing this together with, the following happens:

```
    def test_view(qtbot, df_view, df_model, qtmodeltester):
        qtmodeltester.check(df_model)
        index = df_model.index(2, 0)
        df_view.selectionModel().setCurrentIndex(index, QtCore.QItemSelectionModel.Current)
>       qtbot.stop()
 
test_pandas_model.py:36:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 
self = <pytestqt.qtbot.QtBot object at 0x0000000003EA40A0>
 
    def stopForInteraction(self):
        """
        Stops the current test flow, letting the user interact with any visible widget.
   
        This is mainly useful so that you can verify the current state of the program while writing
        tests.
   
        Closing the windows should resume the test run, with ``qtbot`` attempting to restore visibility
        of the widgets as they were before this call.
   
        .. note:: As a convenience, it is also aliased as `stop`.
        """
        widget_and_visibility = []
        for weak_widget in _iter_widgets(self._request.node):
            widget = weak_widget()
            if widget is not None:
>               widget_and_visibility.append((widget, widget.isVisible()))
E               AttributeError: 'DataFrameModel' object has no attribute 'isVisible'
 
..\..\AppData\Local\Continuum\anaconda3\envs\pyside\lib\site-packages\pytestqt\qtbot.py:273: AttributeError
E
test_pandas_model.py:31 (test_view)
item = <Function test_view>
 
    @pytest.mark.hookwrapper
    @pytest.mark.trylast
    def pytest_runtest_teardown(item):
        """
        Hook called after each test tear down, to process any pending events and
        avoiding leaking events to the next test. Also, if exceptions have
        been captured during fixtures teardown, fail the test.
        """
        _process_events()
>       _close_widgets(item)
 
..\..\AppData\Local\Continuum\anaconda3\envs\pyside\lib\site-packages\pytestqt\plugin.py:181:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 
item = <Function test_view>
 
    def _close_widgets(item):
        """
        Close all widgets registered in the pytest item.
        """
        widgets = getattr(item, "qt_widgets", None)
        if widgets:
            for w, before_close_func in item.qt_widgets:
                w = w()
                if w is not None:
                    if before_close_func is not None:
                        before_close_func(w)
>                   w.close()
E                   AttributeError: 'DataFrameModel' object has no attribute 'close'
 
..\..\AppData\Local\Continuum\anaconda3\envs\pyside\lib\site-packages\pytestqt\qtbot.py:682: AttributeError
                                                  [100%]
 
```

That's a lot of output (actually [even more in reality](https://paste.the-compiler.org/view/6b29f2f4)), and it's difficult to see what's going on there. Thus, I was wondering: Should we maybe have an `isinstance(..., QWidget)` check and bail out early if `addWidget()` is used incorrectly? What do you think?

The-Compiler:
Closes #290

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@
 - ``pytest-qt`` now requires Python 3.6+.
 - ``waitUntil`` now raises a ``TimeoutError`` when a timeout occurs to make the cause of the timeout more explict (`#222`_). Thanks `@karlch`_ for the PR.
 - The ``QtTest::keySequence`` method is now exposed (if available, with Qt >= 5.10).
+- ``addWidget`` now enforces that its argument is a ``QWidget`` in order to display a clearer error when this isn't the case.
 
 .. _#222: https://github.com/pytest-dev/pytest-qt/pull/222
 .. _@karlch: https://github.com/karlch
--- a/src/pytestqt/qtbot.py
+++ b/src/pytestqt/qtbot.py
@@ -166,6 +166,8 @@ class QtBot:
 
         .. note:: This method is also available as ``add_widget`` (pep-8 alias)
         """
+        if not isinstance(widget, qt_api.QWidget):
+            raise TypeError("Need to pass a QWidget to addWidget!")
         _add_widget(self._request.node, widget, before_close_func=before_close_func)
 
     add_widget = addWidget  # pep-8 alias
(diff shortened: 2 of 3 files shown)


nicoddemus:
Awesome, thanks!

PR #302, opened 2020-05-08
RFC: Set application name

The-Compiler:
When an application name is set, Qt uses it internally for e.g. getting
data/config locations:

    >>> QStandardPaths.writableLocation(QStandardPaths.DataLocation)
    '/home/user/.local/share'
    >>> qapp.setApplicationName('pytest-qt-qapp')
    >>> QStandardPaths.writableLocation(QStandardPaths.DataLocation)
    '/home/user/.local/share/pytest-qt-qapp'

That way, when an application accidentally writes into a data/cache/config
directory, it will be easier to find out (rather than those files just landing
somewhere outside an application directory).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@
 - ``waitUntil`` now raises a ``TimeoutError`` when a timeout occurs to make the cause of the timeout more explict (`#222`_). Thanks `@karlch`_ for the PR.
 - The ``QtTest::keySequence`` method is now exposed (if available, with Qt >= 5.10).
 - ``addWidget`` now enforces that its argument is a ``QWidget`` in order to display a clearer error when this isn't the case.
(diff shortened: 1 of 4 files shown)


The-Compiler:
Just noticed a stupid typo after the merge, so I pushed a5ca0f193bd7d96ad5d42ef58d950257bf6889aa to master to fix it. Thanks for the merge! Now I can get rid of overriding the `qapp` fixture in qutebrowser :)

nicoddemus:
Oh sorry I didn't notice that too.

Btw, should we release 4.0 or there are still backward incompatibilities we should include for that major release?

The-Compiler:
I was about to say "drop Qt 4" but then remembered you did that in December :laughing: 

I'll try to take a look over the docs/code and see if I notice anything somewhen in the coming week or so.

nicoddemus:
Awesome, thanks!

PR #308, opened 2020-06-02
Add information for qt 5.15

Czaki:
Qt 5.15 drop bundling `xcb` in binaries so it should be installed locally to perform test under xvfb.


nicoddemus:
Thanks a lot!

PR #326, opened 2020-12-10
Make add_widget no longer an alias, but call addWidget instead

oliveira-mauricio:
We use a subclass of QtBot with a custom `addWidget`. However, when using `add_widget`, which was not redefined in our subclass, it would call the alias in QtBot, calling the parent's addWidget, without our customizations, leading to undesired behavior.

Instead of using an alias, I've added an implementation that works for subclasses as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,9 +9,15 @@
 - ``waitExposed`` and ``waitActive`` now have a default timeout of 5s instead of 1s, in order to match the default
   timeouts Qt uses in the underlying QTest methods.
 - When the ``-s`` (``--capture=no``) argument is passed to pytest, Qt log capturing is now disabled as well.
+- PEP-8 aliases (``add_widget``, ``wait_active``, etc) are no longer just simple
+  assignments to the methods, but they are real methods which call the normal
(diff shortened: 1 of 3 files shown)


nicoddemus (changes requested):
Thanks @oliveira-mauricio!

Currently we have those kind of aliases for all methods, would you mind applying the same change for all of them? I suggest you just group them all in the same part of the file, instead of next to the original methods like today.

```python
# pep-8 aliases

def add_widget(self, *args, **kwargs):
    return self.addWidget(*args, *kwargs)

def wait_active(self, *args, **kwargs):
    return self.waitActive(*args, *kwargs)

def wait_exposed(self, *args, **kwargs):
    return self.waitExposed(*args, *kwargs)

# and so on
```

I also think in those cases we don't really need tests (but thanks for adding one anyway).

nicoddemus (approved):

nicoddemus:
Thanks @oliveira-mauricio!

The-Compiler:
Hmm, I haven't tried yet, but wouldn't this actually break autocompletion in IDEs because they don't know the function signature anymore?

nicoddemus:
You are right, we lose that benefit when using the aliases.

PR #330, opened 2021-01-22
PySide6 support

jensheilman:
Not that many differences to PySide2.
Does not yet support  `from __feature__ import snake_case, true_property`.


src/pytestqt/modeltest.py
```diff
@@ -461,11 +461,11 @@ def _test_data(self):
             (qt_api.QtCore.Qt.SizeHintRole, qt_api.QtCore.QSize),
             (qt_api.QtCore.Qt.FontRole, qt_api.QtGui.QFont),
             (
-                qt_api.QtCore.Qt.BackgroundColorRole,
+                qt_api.QtCore.Qt.BackgroundRole,
                 (qt_api.QtGui.QColor, qt_api.QtGui.QBrush),
             ),
             (
-                qt_api.QtCore.Qt.TextColorRole,
+                qt_api.QtCore.Qt.ForegroundRole,
```
The-Compiler, line 468: I wondered if we should provide backwards-compatibility there, but it looks like Qt's own model tester did the same anyways: https://code.qt.io/cgit/qt/qtbase.git/commit/src/testlib/qabstractitemmodeltester.h?id=e9bebc12812b5c50346952cd9128bdd08ddd0b9d

So I guess it's fine.

src/pytestqt/qt_compat.py
```diff
@@ -110,7 +117,7 @@ def _import_module(module_name):
         self.qInstallMsgHandler = None
         self.qInstallMessageHandler = None
 
-        if self.pytest_qt_api == "pyside2":
+        if self.pytest_qt_api == "pyside2" or self.pytest_qt_api == "pyside6":
```
The-Compiler, line 120: Maybe also set a `self.is_pyside` when setting `self.pytest_qt_api`? Then this doesn't need to be repeated everywhere.

tox.ini
```diff
@@ -1,14 +1,16 @@
 [tox]
-envlist = py{36,37,38}-pyqt5, py{36,37,38}-pyside2, linting
+envlist = py{36,37,38}-pyqt5, py{36,37,38}-pyside2, py{37,38}-pyside6, linting
```
The-Compiler, line 2: Can you please also add test environments for this to CI (`.github/workflows/main.yml`)?

The-Compiler (changes requested):
Thanks! That indeed is a smaller change than what I'd have imagined. I left a couple of comments.

src/pytestqt/modeltest.py
jensheilman: They do have the same values in Qt 5.  https://doc.qt.io/qt-5/qt.html#ItemDataRole-enum

The-Compiler: Yeah, but I was wondering when the new names were introduced - the old ones were deprecated in Qt 5.13, but it looks like the new ones have been around since 4.2 according to https://codereview.qt-project.org/c/qt/qtbase/+/243160. So yup, this certainly isn't an issue.

src/pytestqt/qt_compat.py
jensheilman: changed in  f6397ee.

tox.ini
jensheilman: Changed in 7dd06d7. 
But it seems that only Ubuntu 20.04 is supported by Qt 6. https://www.qt.io/blog/qt6-development-hosts-and-targets

src/pytestqt/qt_compat.py
```diff
@@ -55,23 +56,39 @@ def _can_import(name):
...
         self.pytest_qt_api = self._get_qt_api_from_env() or api or self._guess_qt_api()
+
+        _pyside_modules = [
+            "pyside2",
+            "pyside6",
+        ]
+        if self.pytest_qt_api in _pyside_modules:
+            self.is_pyside = True
+        else:
+            self.is_pyside = False
```
The-Compiler, line 77:
```suggestion
        self.is_pyside = self.pytest_qt_api in ["pyside2", "pyside6"]
```

.github/workflows/main.yml
```diff
@@ -11,7 +11,7 @@ jobs:
 
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
-        qt-lib: [pyqt5, pyside2]
+        qt-lib: [pyqt5, pyside2, pyside6]
         os: [ubuntu-latest, windows-latest, macos-latest]
```
The-Compiler, line 15:
```suggestion
        os: [ubuntu-20.04, windows-latest, macos-latest]
```

unless @nicoddemus sees a reason why we should keep using Ubuntu 18.04 for the other environments - but "ubuntu-latest" will soon be 20.04 anyways: https://github.com/actions/virtual-environments/issues/1816

src/pytestqt/qtbot.py
```diff
@@ -191,7 +191,7 @@ def waitActive(self, widget, timeout=5000):
 
         .. note::
             This function is only available in PyQt5, raising a ``RuntimeError`` if called from
-            ``PySide2``.
+            ``PySide2/6`.
```
The-Compiler, line 194:
```suggestion
            ``PySide2/6``.
```

The-Compiler (changes requested):

The-Compiler (approved):
Looking great now! :tada: Thanks for the follow-ups! I'll wait for a bit to see what @nicoddemus says about it.

nicoddemus (approved):
LGTM, thanks a lot @jensheilman for the excellent PR and @The-Compiler for the extensive review. 🎉

nicoddemus:
Should we squash/merge this?

nicoddemus:
Ahh we also need to update the CHANGELOG, will do it now.

nicoddemus (approved):

j9ac9k:
Posting here in case it's helpful, we rolling out PySide6 and PyQt6 Compatibility in pyqtgraph recently; PyQt6 required much more work... you can see the PRs here:

PySide6: https://github.com/pyqtgraph/pyqtgraph/pull/1495
PyQt6: https://github.com/pyqtgraph/pyqtgraph/pull/1502

nicoddemus:
Thanks @j9ac9k!

@The-Compiler should we go ahead and merge this?

The-Compiler:
Looks good to me! :+1: Not sure if you prefer to squash or not, so I'll let you press the button :wink:

The-Compiler:
FYI, this finally [got released](https://twitter.com/the_compiler/status/1400490856549949452) as part of pytest 4.0.0! :tada: 

Thanks again @jensheilman for the initial Qt6 and PySide6 support! :+1:

Issue #333, opened 2021-02-19
Dynamically generated QTest methods not friendly to type checkers and IDEs


TilmanK:
When writing tests with QtBot (which is a blessing!) I also annotate my tests to easily find errors when writing them.

As an example a test definition looks like this:
def test_filtering(qtbot: QtBot) -> None:

I've silenced mypy for pytest-qt anyway, but PyCharm keeps complaining about QtBot missing methods (that originate from the QTest api).

From my point of view (which might be limited and not complete) it would be sufficient to add definitions to QtBot to tell PyCharm (and other IDEs?) what to expect:

def mouseClick(self, button, mouse_button, stateKey=0, pos=None, delay=-1):...

I'd send a PR for this, if ok.

nicoddemus:
Hi @TilmanK,

Indeed right now QTest methods are injected dynamically:

https://github.com/pytest-dev/pytest-qt/blob/f3ec8653357bb3de886c2b36ff5be243955f11db/src/pytestqt/qtbot.py#L600-L642

It would be great to define them explicitly to help type checkers and IDEs. 👍 

(I've updated the description, hope you don't mind)

TilmanK:
Great @nicoddemus!

Is the definition as mentioned above sufficient? Or do we need something else?

nicoddemus:
Should be enough (of course for all methods plus calling the actual implementation). 👍

TilmanK:
We could do that, but with the implementation as it is today, the method definitions are overwritten at runtime (as far as I understand) by the existing code. Is there any benefit in calling them from the definitions to come?

nicoddemus:
I think the approach is to replace the current implementation, which creates proxy methods:

```python
# inject methods from QTest into QtBot 
method_names = [ 
    "keyPress",
    "mouseClick", 
    ...
]
if hasattr(qt_api.QtTest.QTest, "keySequence"): 
    # Added in Qt 5.10 
    method_names.append("keySequence") 

for method_name in method_names: 
    method = create_qtest_proxy_method(method_name) 
    if method is not None: 
        setattr(cls, method_name, method)      
```

By explicit calls:

```python
@staticmethod
def keyPress(widget, key, modifier=0, delay=-1) -> None:
    qt_api.QtTest.QTest.keyPress(widget, key, modifier, delay)

@staticmethod
def mouseClick(widget, mouse_button, stateKey=0, pos=None, delay=-1) -> None:
[...]

PR #334, opened 2021-02-21
Explicit definition of QTest API methods. #333

TilmanK:
Removes the injection of QTest API methods at runtime and defines them explicitly. Adds a test that verifies that calling keyToAscii on PyQt isn't possible (because not available).

### DIFF
--- a/src/pytestqt/plugin.py
+++ b/src/pytestqt/plugin.py
@@ -226,10 +226,6 @@ def pytest_configure(config):
             DeprecationWarning,
         )
 
-    from .qtbot import QtBot
-
-    QtBot._inject_qtest_methods()
-
 
 def pytest_report_header():
     from pytestqt.qt_compat import qt_api
--- a/src/pytestqt/qtbot.py
+++ b/src/pytestqt/qtbot.py
@@ -1,5 +1,4 @@
 import contextlib
-import functools
 import weakref
 
 from pytestqt.exceptions import SignalTimeoutError, TimeoutError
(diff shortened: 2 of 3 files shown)


TilmanK:
@nicoddemus: Can you advise how to test that keySequence isn't working (raises) on qt versions <= 5.10? I don't have a clue (my pytest knowledge isn't enough)...

nicoddemus (approved):
LGTM, thanks a lot, great work!

ferdnyc:
@TilmanK
> Can you advise how to test that keySequence isn't working (raises) on qt versions <= 5.10?

I'm not @nicoddemus so it's possible I missed something, but I don't think there is a way. Looking at [the CI runs](https://github.com/pytest-dev/pytest-qt/actions/runs/589899971), unlike the matrix of Python versions and Qt interfaces, there's no spread of Qt versions ­— all of the tests are run with the same Qt version (5.15.2, currently). Understandable, since adding others would be... a chore, and for little real benefit.

To write that test you'd need older versions of both Qt and PyQt5. (PySide2 didn't even exist until Qt 5.11, so it's a moot point there.) Older PyQt5 requires older sip packages, and the wheels for [compatible sip versions](https://pypi.org/project/sip/4.19.3/#files) are only built for Python 3.5 and 3.6. Try to install PyQt5 5.9 in Python 3.9, and you hit this:
```sh
(venv)$ pip3 install pyqt5==5.9
Collecting pyqt5==5.9
  Downloading PyQt5-5.9-5.9.1-cp35.cp36.cp37-abi3-manylinux1_x86_64.whl (104.6 MB)
     |████████████████████████████████| 104.6 MB 14 kB/s 
ERROR: Could not find a version that satisfies the requirement sip<4.20,>=4.19.3 (from pyqt5)
ERROR: No matching distribution found for sip<4.20,>=4.19.3
(venv)$
```

(It is technically doable in Python 3.6; there you can `pip3.6 install pyqt5==5.9` successfully, at which point this works):
```python3
Python 3.6.13 (default, Feb 16 2021, 00:00:00) 
[GCC 10.2.1 20201125 (Red Hat 10.2.1-9)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pytestqt.qt_compat import qt_api
>>> qt_api.set_qt_api('pyqt5')
>>> qt_api.get_versions().qt_api_version
'5.9'
```

_Doable?_ Yes. _Worth it?_ Highly doubtful, IMHO.

nicoddemus:
Ahh sorry, I missed the question: I misunderstood it as asking for help because the build was failing and was later fixed, not that @TilmanK wanted to ask the actual failure.

But I completely agree with @ferdnyc, thanks for the detailed response. 👍

TilmanK:
@ferdnyc Yes, you're right. I haven't thought about the usefulness of that test.

PR #337, opened 2021-03-16
Clean up remaining Qt 4 compat code

The-Compiler:
I found various code snippets which don't seem to be needed anymore since Qt 4 is dropped.

### DIFF
--- a/src/pytestqt/modeltest.py
+++ b/src/pytestqt/modeltest.py
@@ -78,7 +78,7 @@ class ModelTester:
             return "{}/{} {!r} (0x{:x})".format(
                 index.row(),
                 index.column(),
-                qt_api.extract_from_variant(data),
+                data,
                 id(index),
             )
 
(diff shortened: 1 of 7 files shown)


src/pytestqt/modeltest.py
```diff
@@ -484,14 +484,14 @@ def _test_data(self):
         for role, typ in types:
             data = self._model.data(self._model.index(0, 0), role)
             if data is not None:
-                data = qt_api.extract_from_variant(data)
+                data = data
```
nicoddemus, line 487: This line and the `if` above can be removed.

src/pytestqt/modeltest.py
```diff
@@ -484,14 +484,14 @@ def _test_data(self):
...
             assert data == None or isinstance(data, typ), role  # noqa
 
         # Check that the alignment is one we know about
         alignment = self._model.data(
             self._model.index(0, 0), qt_api.QtCore.Qt.TextAlignmentRole
         )
-        alignment = qt_api.extract_from_variant(alignment)
+        alignment = alignment
```
nicoddemus, line 494: This line can be removed.

nicoddemus (approved):
Cool, thanks!

The pyside/Qt 6 failures are unrelated, feel free to merge even with the failures.

The-Compiler:
I'm working on a fix for the Qt 6 test issues. Yep, I agree this can be cleaned up further, missed that with a search/replace :sweat_smile: 

I even wonder if we can drop `make_variant` as well, as typically with PyQt5 you don't need to create a `QVariant` explicitly. I'll take a look.

nicoddemus (approved):

Issue #338, opened 2021-03-16
filterwarnings does not seem to work


The-Compiler:
In `tests/test_wait_signal.py`, we have:

```python
@pytest.mark.filterwarnings("ignore:qt_wait_signal_raising is deprecated")
@pytest.mark.parametrize("configkey", ["qt_wait_signal_raising", "qt_default_raising"])
def test_raising_by_default_overridden(qtbot, testdir, configkey):
```

yet running pytest results in:

```

=============================== warnings summary ===============================
tests/test_wait_signal.py::test_raising_by_default_overridden[qt_wait_signal_raising]
  /home/florian/proj/pytest-qt/.tox/py37-pyside6/lib/python3.7/site-packages/pytestqt/plugin.py:226: DeprecationWarning: qt_wait_signal_raising is deprecated, use qt_default_raising instead.
    DeprecationWarning,

-- Docs: https://docs.pytest.org/en/stable/warnings.html
================== 206 passed, 3 skipped, 1 warning in 23.79s ==================
```

any idea why? I'm stumped...

nicoddemus:
Sorry for the delay.

I suspect this is because some early warnings cannot really be ignored by pytest's filterwarnings configuration (because they are not captured). I will see if I can investigate this when I have some time to give a definitive answer.

PR #339, opened 2021-03-16
Fix Qt 6 test failures

The-Compiler:
Looks like the build [passed](https://github.com/pytest-dev/pytest-qt/runs/1764317110) in the PR (#330) but then [started failing](https://github.com/pytest-dev/pytest-qt/runs/2008980207) with the merge to master (996b9d617fd3585e43693744bb851acbed0fd1df).

The former used PySide6 6.0.0 while the latter used PySide6 6.0.1 - I suspect that broke those tests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ universal = 1
 
 [tool:pytest]
 testpaths = tests
-addopts = --strict
+addopts = --strict-markers --strict-config
 xfail_strict = true
 markers =
   filterwarnings: pytest's filterwarnings marker
--- a/src/pytestqt/qtbot.py
+++ b/src/pytestqt/qtbot.py
@@ -623,7 +623,7 @@ class QtBot:
 
     @staticmethod
     def keyToAscii(key):
-        if qt_api.pytest_qt_api == "pyqt5":
+        if not hasattr(qt_api.QtTest.QTest, "keyToAscii"):
             raise NotImplementedError("This method isn't available on PyQt5.")
         qt_api.QtTest.QTest.keyToAscii(key)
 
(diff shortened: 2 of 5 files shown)


src/pytestqt/qtbot.py
```diff
@@ -623,7 +623,7 @@ def keySequence(widget, key_sequence):
 
     @staticmethod
     def keyToAscii(key):
-        if qt_api.pytest_qt_api == "pyqt5":
+        if not hasattr(qt_api.QtTest.QTest, "keyToAscii"):
```
The-Compiler, line 626: Seems like the better check to me, given that `keyToAscii` apparently wasn't available with PySide6 6.0.0 (otherwise the test would've failed in #330 already).

tox.ini
```diff
@@ -15,7 +15,7 @@ setenv=
     pyqt5: PYTEST_QT_API=pyqt5
     QT_QPA_PLATFORM=offscreen
 
-passenv=DISPLAY XAUTHORITY USER USERNAME
+passenv=DISPLAY XAUTHORITY USER USERNAME COLUMNS
```
The-Compiler, line 18: Unrelated, but I typically use `COLUMN=80 tox -e ...` when I want copyable pytest-output, like for #338.

setup.cfg
```diff
@@ -3,7 +3,7 @@ universal = 1
 
 [tool:pytest]
 testpaths = tests
-addopts = --strict
+addopts = --strict-markers --strict-config
```
The-Compiler, line 6: To avoid pytest's deprecation warning.

The-Compiler:
cc @jensheilman :slightly_smiling_face:

nicoddemus:
Thanks @The-Compiler! I've been meaning to get to this, but time has been short lately. 🙇

nicoddemus (approved):

Issue #306, opened 2020-05-25
API break suggestions for 4.0


The-Compiler:
As mentioned in https://github.com/pytest-dev/pytest-qt/pull/302#issuecomment-629513007 I went through the docs/code and thought about possible API breaks which would make sense for 4.0. Here are a couple of ideas:

- [x] 1. One thing that always irked me a bit is that we have two naming conventions (`add_widget` vs. `addWidget`). Is it really a good idea to support both? At least in my project, it only lead to using both inconsistently over the years without noticing. Should we get rid of one of those? On the other hand, that's probably a lot of churn for projects using pytest-qt, so I'm not sure if it's worth it. Maybe as an alternative, we could make it possible to configure the desired API style in `pytest.ini` so that the other style would raise an exception (that could be done after 4.0)?
- [x] 2. Right now the user needs to decide between `waitActive`, `waitExposed` and `waitForWindowShown` (which might be a hurdle if using the same tests for two different backends). Should we have a wrapper which decides what to use instead?
- [x] 3. Shouldn't `waitForWindowShown` have a `timeout` as well? (can be done after 4.0)
- [x] 4. Does anyone really use `qtbot.stopForInteraction()` instead of `qtbot.stop()`? A GitHub search [reveals 60 results](https://github.com/search?q=qtbot.stopForInteraction&type=Code), vs. [122 results](https://github.com/search?q=%22qtbot.stop%22&type=Code) for `.stop()`, though I guess it's mostly used interactively for debugging without being added to the code. Should we remove the long name?
- [x] 5. Should we raise the default timeout for `waitSignal()` to 5000, similar to [QSignalSpy.wait()](https://doc.qt.io/qt-5/qsignalspy.html#wait) in Qt and the other default timeouts in pytest-qt?
- [x] 6. `waitSignal` and `waitSignals` can take `None` instead of a signal, in which case they just wait until the timeout occurred. But we already have `qtbot.wait()` for that, so is there a reason to support it?
- [x] 7. Can we get rid of the `SignalTimeoutError` compatibility alias for `TimeoutError`?
- [x] 8. The docs for `pytestqt.logging.Record` says that `context` is only available for Qt 5, that can probably be removed now (can be done after 4.0)
- [x] 9. (new, see below) Get rid of `qt_wait_signal_raising`
- [x] 10. (new, see below) Get rid of aliases in `plugin.py`

nicoddemus:
Thanks a lot for working on this list Florian!

[...]

PR #340, opened 2021-03-16
Support PyQt6

The-Compiler:
Still a couple of issues (marked as FIXME for now), but wanted to get some feedback.

cc @jensheilman

Note that with the scoped enum access, we're dropping support for PyQt5 < 5.11. I think that's fine, given that 5.11 has been released in July 2018. I haven't checked yet how far back scoped enum access support in PySide2 goes, though.

### DIFF
--- a/src/pytestqt/qt_compat.py
+++ b/src/pytestqt/qt_compat.py
@@ -84,6 +84,8 @@ class _QtApi:
             )
             raise RuntimeError(msg)
 
+        # FIXME check minimum supported versions?
+
         _root_modules = {
             "pyside6": "PySide6",
             "pyside2": "PySide2",
@@ -168,17 +170,7 @@ class _QtApi:
             assert self.is_pyqt
             self.isdeleted = _import_module("sip").isdeleted
 
-        # PyQt6 only allows flag access via the flag type (usually plural),
-        # while PyQt5/PySide2/PySide6 only allow such access via the enum type
-        # (usually singular):
-        # https://www.riverbankcomputing.com/pipermail/pyqt/2021-March/043709.html
         if self.pytest_qt_api == "pyqt6":
-            self.AlignmentFlag = self.Qt.Alignment
-            self.Orientation = self.Qt.Orientations
-            self.MouseButton = self.Qt.MouseButtons
-            self.KeyboardModifier = self.Qt.KeyboardModifiers
-            self.ItemFlag = self.Qt.ItemFlags
-
             # exec was a keyword in Python 2, PyQt6 dropped the .exec_ alias but
             # PySide2/PySide6 still name it "exec_" only.
             self.QtCore.QCoreApplication.exec_ = (
@@ -190,12 +182,6 @@ class _QtApi:
             _QtWidgets.QDialog.exec_ = lambda _self, *args, **kwargs: _self.exec(
                 *args, **kwargs
             )
-        else:
-            self.AlignmentFlag = self.Qt.AlignmentFlag
-            self.Orientation = self.Qt.Orientation
-            self.MouseButton = self.Qt.MouseButton
-            self.KeyboardModifier = self.Qt.KeyboardModifier
-            self.ItemFlag = self.Qt.ItemFlag
 
     def get_versions(self):
         if self.pytest_qt_api == "pyside6":
--- a/src/pytestqt/qt_compat.py
+++ b/src/pytestqt/qt_compat.py
@@ -209,7 +209,7 @@ class _QtApi:
                 self.QtCore.QT_VERSION_STR,
             )
 
-        assert False, self.pytest_qt_api
+        assert False, f"Internal error, unknown pytest_qt_api: {self.pytest_qt_api}"
 
 
 qt_api = _QtApi()
--- a/src/pytestqt/qt_compat.py
+++ b/src/pytestqt/qt_compat.py
@@ -39,6 +39,7 @@ class _QtApi:
             if api not in (
                 "pyside6",
                 "pyside2",
+                "pyqt6",
                 "pyqt5",
             ):  # pragma: no cover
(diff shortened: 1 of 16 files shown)


The-Compiler:
I think the segfaults/aborts are all because of the same issue in PyQt6, which I reported here: [PyQt6: Double free in sipWrapper_dealloc with QApplication.postEvent](https://www.riverbankcomputing.com/pipermail/pyqt/2021-March/043711.html)

nicoddemus:
Thanks for looking into this!

> I think the segfaults/aborts are all because of the same issue in PyQt6, which I reported here: PyQt6: Double free in sipWrapper_dealloc with QApplication.postEvent

Skipping those tests sounds good, perhaps just add that link to the `reason`?

The-Compiler:
I've just verified that the oldest PySide2 release (5.11.0) works with the scoped enum access (e.g. `QtCore.Qt.Orientation.Horizontal` instead of `QtCore.Qt.Horizontal`. Thus, this change only drops support for PyQt5 < 5.11. I think that's fine given its age and that the next release will be a major one, would you agree?

As for the aliases for flags and the crashing tests, it looks like both are going to be fixed in PyQt 6.1 - would you prefer getting this in before that, or should we wait (probably around a month) so we can remove those workarounds?

nicoddemus:
> I think that's fine given its age and that the next release will be a major one, would you agree?

Agreed.

> would you prefer getting this in before that, or should we wait (probably around a month) so we can remove those workarounds?

I think we can wait, what do you say?

The-Compiler:
I just updated this to install the PyQt6 6.1 beta from [Riverbank's PyPI server](https://www.riverbankcomputing.com/pypi/) - much simpler now as the flags were renamed, and the testsuite now runs fine as well.

The only remaining issue from my point of view is the "FIXME check minimum supported versions?" - I think it'd be good if we're on unsupported PyQt versions (PyQt5 < 5.11 and PyQt 6.0) and show a proper error. Other than that, this should be ready for another review.

I'm a bit short on time right now, likely won't get back to this in the next 2 weeks. Given that the Qt 6.1 release [is planned](https://wiki.qt.io/Qt_6.1_Release) for 27th and PyQt releases usually follow shortly afterwards, I assume we can probably drop the `commands_pre` installing the snapshot before this is merged.

nicoddemus:
Thanks @The-Compiler!

I will review this next week then. 👍

src/pytestqt/qt_compat.py
```diff
@@ -153,6 +155,33 @@ def _import_module(module_name):
...
+        elif self.pytest_qt_api == "pyside6":
+            import shiboken6
+
+            self.isdeleted = lambda obj: not shiboken6.isValid(obj)
+        else:
+            assert self.is_pyqt
+            self.isdeleted = _import_module("sip").isdeleted
+
+        if self.pytest_qt_api == "pyqt6":
+            # exec was a keyword in Python 2, PyQt6 dropped the .exec_ alias but
```
nicoddemus, line 174: Not sure we should leak out those names to user space... perhaps better to have `self.QCoreApplication_exec = (...` and we use that internally? Same for the other two.

src/pytestqt/qt_compat.py
```diff
@@ -172,13 +200,22 @@ def get_versions(self):
...
+            )
+        elif self.pytest_qt_api == "pyqt5":
             return VersionTuple(
                 "PyQt5",
                 self.QtCore.PYQT_VERSION_STR,
                 self.QtCore.qVersion(),
                 self.QtCore.QT_VERSION_STR,
             )
 
+        assert False, self.pytest_qt_api
```
nicoddemus, line 218:
```suggestion
        assert False, f"Internal error, unknown pytest_qt_api: {self.pytest_qt_api}"
```

nicoddemus: Just in case this somehow happens to a user.

nicoddemus (changes requested):
Great work @The-Compiler, thanks a lot for tackling this!

I left a few minor comments, and also I agree with you that we should check the supported versions to properly error out. 

As you said, we can wait `6.1` to come out (scheduled for next week?) to merge this. 👍

The-Compiler:
> As you said, we can wait `6.1` to come out (scheduled for next week?) to merge this.

The RC came out on Thursday [as planned](https://wiki.qt.io/Qt_6.1_Release), so I'm assuming the final will follow as planned on the 6th of May (so more like 1.5 weeks). That's actually pretty good timing as I'm giving another pytest training next week (this month has been crazy!) so I'll have time to finish this in early May.

Next I want to pick up #306 so we can finally do another release (the last one was in late 2019).

src/pytestqt/qt_compat.py
The-Compiler: Good point, I didn't consider that! Yeah, let's go for a function instead then.

nicoddemus:
> That's actually pretty good timing as I'm giving another pytest training next week (this month has been crazy!) so I'll have time to finish this in early May.

Sounds great! Have a great training. 👍 

> Next I want to pick up #306 so we can finally do another release (the last one was in late 2019).

Cool, thanks!

The-Compiler:
Quick update: Qt 6.1 was [released](https://www.qt.io/blog/qt-6.1-released) now, I guess PyQt will follow in the next couple of days, at which point I'll pick this up again.

nicoddemus (approved):
Thanks! Reviewed the last commits individually. 👍 

Looks great, ship it! 🚀

tlambert03:
Thanks for this!  have been using your in-progress branch for some CI tests for a while now and it's been working great.  Great to have this in :)

The-Compiler:
@tlambert03 Nice, thanks for the feedback. 4.0.0 [is out now](https://twitter.com/the_compiler/status/1400490856549949452)! :tada:

PR #349, opened 2021-05-08
Force garbage collection after each test

bartfeenstra:
This solves the problem that widgets opened during a test are kept alive in the global QApplication instance, which can be tested by `assert []==qapp.topLevelWidgets()` at the beginning of the second, third, etc. test after the first test opens widgets.

It does require the functional change to the `qapp` fixture scope from `session` to `function`. I don't know if that would be considered breaking or not.

### DIFF
--- a/src/pytestqt/plugin.py
+++ b/src/pytestqt/plugin.py
@@ -1,3 +1,4 @@
+import gc
 import warnings
 
 import pytest
@@ -42,7 +43,7 @@ def qapp_args():
     return []
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
 def qapp(qapp_args, pytestconfig):
     """
     Fixture that instantiates the QApplication instance that will be used by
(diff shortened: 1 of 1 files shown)


bartfeenstra:
Test coverage is forthcoming.

nicoddemus:
> FWIW I don't think this should be part of the qapp fixture - why not make it a separate fixture?

I agree, something like this:

```python
@pytest.fixture(autouse=True)
def run_gc():
    yield
    gc.collect()
```

Thanks @bartfeenstra for the PR anyway, we appreciate taking the time.

bartfeenstra:
Thanks for taking the time to write up a considerate and useful conclusion :heart:  The tests took me more effort than expected (other parts of the code base failing) and I hadn't had the time to step through that this week yet. I did add a custom fixture similar to the one I added in this PR to my project and that works fine for me.

nicoddemus:
Yeah I know how hard it is to track leak-related failures in a suite, hehehe. Good luck, and thanks again!

PR #356, opened 2021-05-27
Improve qt_compat error messages

### LINKED ISSUE #355: Use pytest.UsageError instead of RuntimeError for qt_compat errors
For user errors (e.g. wrong `PYTEST_QT_API` value or too old PyQt version) we should use `pytest.UsageError` instead of `RuntimeError` so that the user gets a nice error message instead of an `INTERNALERROR>` traceback.

Will do this after #340 since this adds two new such errors.

The-Compiler:
Fixes #355


nicoddemus (approved):
Thanks!

PR #358, opened 2021-05-27
Clean up old qt_compat aliases

### LINKED ISSUE #354: More qt_compat cleanups
I think there's more clean-up work we can do in qt_compat after #340 is in.

The-Compiler:
Removes all aliases to class names, since those were only needed back when those
classes were in different modules between Qt 4 and Qt 5 (typically QtGui vs.
QtWidgets).

With Qt 4 now dropped, Qt 5 and Qt 6 are compatible enough to use the full path
to the classes inside our code instead, which makes the wrapper much simpler.

This also cleans up some tests doing "QtCore = qt_api.QtCore" or similar, since
there were a lot of duplicated but unused such aliases. Seems simpler to me to
just use the direct name, even if a bit longer.

Fixes #354

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,9 @@
   implementations. This makes subclasses work as expected, instead of having to
   duplicate the assignment. Thanks `@oliveira-mauricio`_ for the PR.
 - The ``qt_api.extract_from_variant`` and ``qt_api.make_variant`` functions
-  (which were never intended for public usage) are now removed.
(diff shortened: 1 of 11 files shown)


nicoddemus (approved):

The-Compiler:
Wheee, thanks for all the quick reviews!

nicoddemus:
> Wheee, thanks for all the quick reviews!

Sure thing, thanks for tackling all this and spearheading the next release. 😁

PR #359, opened 2021-05-27
Initial API cleanups

The-Compiler:
See #306. I reverted the commit regarding the imports in plugin.py, as the rest should be uncontroversial. Let's look at that one separately.

### DIFF
--- a/src/pytestqt/exceptions.py
+++ b/src/pytestqt/exceptions.py
@@ -98,14 +98,6 @@ class TimeoutError(Exception):
     .. versionadded:: 2.1
 
     Exception thrown by :class:`pytestqt.qtbot.QtBot` methods.
-
-    .. note::
-        In versions prior to ``2.1``, this exception was called ``SignalTimeoutError``.
-        An alias is kept for backward compatibility.
     """
 
     pass
-
-
-# backward compatibility alias
-SignalTimeoutError = TimeoutError
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -47,6 +47,8 @@
   with the existing ``waitActive`` and ``waitExposed`` timeouts, as well as the
   default timeout used by Qt's ``QSignalSpy::wait``. To get the old behavior
   back, explicitly pass ``timeout=1000`` to those functions.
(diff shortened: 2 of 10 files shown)


src/pytestqt/exceptions.py
```diff
@@ -98,14 +98,6 @@ class TimeoutError(Exception):
...
-    .. note::
-        In versions prior to ``2.1``, this exception was called ``SignalTimeoutError``.
-        An alias is kept for backward compatibility.
     """
 
     pass
-
-
-# backward compatibility alias
-SignalTimeoutError = TimeoutError
```
nicoddemus, line 111: I was wondering this days if it wouldn't make sense to ditch our custom `TimeoutError` and raise the builtin `TimeoutError` instead... what do you think?

nicoddemus (approved):
LGTM, thanks!

Leaving a small comment to see what you think. 👍

src/pytestqt/exceptions.py
The-Compiler: Python's `TimeoutError` is a `OSError` subclass, and documented as:

> Raised when a system function timed out at the system level. Corresponds to errno `ETIMEDOUT`.

So that doesn't really seem like a good match for us. Additionally, this would cause [more churn than I'm comfortable with](https://github.com/search?q=pytestqt.exceptions+-extension%3Ajson+-extension%3Aebuild&type=Code) for something we already changed one time before, so I don't think it's worth it.

The-Compiler:
I rebased to remove the revert of the compatibility API deletion, based on https://github.com/pytest-dev/pytest-qt/issues/306#issuecomment-849844692.

The only thing left is how to handle `waitActive` / `waitExposed` / `waitForWindowShown` then. I'll need to think some more about this, probably tomorrow.

src/pytestqt/exceptions.py
nicoddemus: Fair enough. 👍

The-Compiler:
Oh, I just noticed using `pytestqt.plugin.QtApi` will actually still be possible, as we need the import for the `qtbot` fixture in there. So this should really only affect qutebrowser which imports `SignalBlocker` from there :sweat_smile:

PR #363, opened 2021-06-02
Update docs for 4.0

The-Compiler:
Almost done now! Unfortunately I noticed another thing to look at :laughing:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,63 +1,88 @@
 4.0.0 (UNRELEASED)
 ------------------
 
-- `PySide6 <https://pypi.org/project/PySide6>`__ is now supported. Thanks `@jensheilman`_ for the PR.
-- `PyQt6 <https://pypi.org/project/PyQt6>`__ 6.1+ is now supported. Thanks `@The-Compiler`_ for the PR.
+- `PySide6 <https://pypi.org/project/PySide6>`__ and `PyQt6 <https://pypi.org/project/PyQt6>`__ (6.1+)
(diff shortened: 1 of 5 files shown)


nicoddemus (approved):
Great!

CHANGELOG.rst
```diff
@@ -1,63 +1,88 @@
...
+  removed. Thanks `@The-Compiler`_ for the PR.
 - The default timeouts for ``qtbot.waitSignal``, ``waitSignals``, ``waitUntil``
-  and ``waitCallback`` have been raised from 1s to 5s. This makes them in line
-  with the existing ``waitActive`` and ``waitExposed`` timeouts, as well as the
-  default timeout used by Qt's ``QSignalSpy::wait``. To get the old behavior
-  back, explicitly pass ``timeout=1000`` to those functions.
+  and ``waitCallback``, ``waitActive`` and ``waitExposed`` have been raised from
+  1s to 5s. This makes them in line the default timeout used by Qt's underlying
+  methods such as ``QSignalSpy::wait``. To get the old behavior back, explicitly
+  pass ``timeout=1000`` to those functions (`#306`_). Thanks `@The-Compiler`_
```
nicoddemus, line 49: While reading this I realized `waitSignal` (and friends) have this signature:

```python
def waitSignal(self, signal, timeout=5000, raising=None, check_params_cb=None):
```

While IMHO this would be better:

```python
def waitSignal(self, signal, *, timeout=5000, raising=None, check_params_cb=None):
```

(Of course this is an artifact from Python 2 support).

Don't want to pile this on for 4.0, just commenting to see what you think.

The-Compiler: Let's do this for 4.0! See #366. Makes total sense, and should be an easy change.

PR #375, opened 2021-06-13
Change duplicate PyQt6 to PySide6

mara004:


nicoddemus (approved):
Thanks for the catch!

The-Compiler:
Whoops, thanks! We had another one in the README, fixed in 5e57199da66594986f81d8cb1b07c18aefa68f5a - unfortunately I only noticed after the 4.0.2 release was out already :sweat_smile:

PR #383, opened 2021-08-04
Simplify PEP8 method aliases in QtBot class.

luziferius:
This commit simplifies the creation of PEP8 compliant method aliases.

It reduces call stack depth by one and enables better parameter suggestions in IDEs.

Before:
qtbot_args

After:
qtbot_args_new

Also added unit test verifying that each alias points to the correct source method.

(I hope that the test function location is OK)

### DIFF
--- a/src/pytestqt/qtbot.py
+++ b/src/pytestqt/qtbot.py
@@ -138,6 +138,17 @@ class QtBot:
 
     def __init__(self, request):
         self._request = request
+        # pep8 aliases. Set here to automatically use implementations defined in sub-classes for alias creation
+        self.add_widget = self.addWidget
+        self.capture_exceptions = self.captureExceptions
+        self.wait_active = self.waitActive
+        self.wait_exposed = self.waitExposed
+        self.wait_for_window_shown = self.waitForWindowShown
+        self.wait_signal = self.waitSignal
+        self.wait_signals = self.waitSignals
+        self.assert_not_emitted = self.assertNotEmitted
+        self.wait_until = self.waitUntil
+        self.wait_callback = self.waitCallback
 
     def _should_raise(self, raising_arg):
         ini_val = self._request.config.getini("qt_default_raising")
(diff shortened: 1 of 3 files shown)


nicoddemus:
Thanks @luziferius for the PR!

However that way to declare the aliases was how it was done some time ago, until https://github.com/pytest-dev/pytest-qt/pull/326. Please take a look there to see the rationale and let us know what you think.

luziferius:
> Thanks @luziferius for the PR!
> 
> However that way to declare the aliases was how it was done some time ago, until #326. Please take a look there to see the rationale and let us know what you think.

Oops. Haven’t seen that. I just assumed it was a legacy implementation that was this way because of Python 2.x compatibility.

I’ll experiment with another solution then.

luziferius:
I think I have found a solution.

Are you fine with a solution that breaks, if the subclass overwrites `__init__()` but does not perform a `super()` call in it’s own `__init__()` implementation?

I’d consider doing so an illegal use of the base class, so it should be fine from my point of view.

src/pytestqt/qtbot.py
```diff
@@ -646,33 +656,15 @@ def mouseRelease(*args, **kwargs):
...
-
-    def assert_not_emitted(self, *args, **kwargs):
-        return self.assertNotEmitted(*args, **kwargs)
-
-    def wait_until(self, *args, **kwargs):
-        return self.waitUntil(*args, **kwargs)
-
-    def wait_callback(self, *args, **kwargs):
-        return self.waitCallback(*args, **kwargs)
+    add_widget = addWidget
```
nicoddemus, line 659: Isn't this redundant with what's done in `__init__`?

nicoddemus:
Thanks!

> I’d consider doing so an illegal use of the base class, so it should be fine from my point of view.

I agree, that would be fine.

src/pytestqt/qtbot.py
luziferius: Technically, it is.

The static assignments allow analysis tools to determine that there is an alias and provide auto-completions in IDEs. If the analysis tool can handle the assignment in `__init__()`, the definitions in the class body can be removed.

The re-assignment inside __init__() overwrites these with potential re-implementations in sub-classes.

luziferius: Ok, I tested it in PyCharm and it works there without the definition in the class body.

So should I remove the aliases there and only keep them in `__init__()`?

If yes, what’s your commit policy? Add another commit or squash and force-push?

luziferius: I removed the redundant definitions

luziferius:
Done. The PEP 8 aliases are now only defined in `__init__()`.

The unit tests added do:

- Both tests auto-discover PEP8 aliases, to make sure that aliases added in the future will be covered as well.
- Check that the alias point to the right method, i.e. that assignments are not shuffled around, like the faulty `self.wait_active = self.waitSignal`. This also prevents accidentally adding an independent method that just looks like a PEP8 alias.
- Check that the PEP8 alias points to the correct implementation when sub-classing QtBot. The test programmatically generates subclasses that overwrite a camelCaseMethod and verifies that the alias points to the overwritten method.

nicoddemus:
Thanks @luziferius!

The PR and the solution is very well written, and the added checks are the cherry on the cake. Thanks! 🍰 

My very (small) concern is that we are creating a reference cycle in `QtBot.__init__`, but I don't think it will really be an issue given `QtBot` is small in size.

nicoddemus (approved):

luziferius:
> My very (small) concern is that we are creating a reference cycle in `QtBot.__init__`, but I don't think it will really be an issue given `QtBot` is small in size.

Would you like to elaborate on this?

Do you mean something like this below or something entirely different?
``` python
self.a = self.b
self.b = self.a
```

nicoddemus:
Yeah I mean:

```
self.add_widget = self.addWidget
```

Creates a circular reference because `self.add_widget` is a bound method to `self.addWidget`, which references `self`.

PR #412, opened 2022-03-06
Check if backend is loaded

eyllanesc:
- [x] tests.
- [ ] CHANGELOG.

### DIFF
--- a/src/pytestqt/qt_compat.py
+++ b/src/pytestqt/qt_compat.py
@@ -11,6 +11,7 @@ Based on from https://github.com/epage/PythonUtils.
 
 from collections import namedtuple
 import os
+import sys
 
 import pytest
 
@@ -21,6 +22,10 @@ VersionTuple = namedtuple("VersionTuple", "qt_api, qt_api_version, runtime, comp
 def _import(name):
     """Think call so we can mock it during testing"""
     return __import__(name)
+
+
+def _is_library_loaded(name):
+    return name in sys.modules
 
 
 class _QtApi:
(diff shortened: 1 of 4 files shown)


src/pytestqt/qt_compat.py
```diff
@@ -50,6 +55,17 @@ def _get_qt_api_from_env(self):
                 raise pytest.UsageError(msg)
         return api
 
+    def _get_backend_loaded(self):
```
nicoddemus, line 58: How about `_get_already_loaded_backend`?

nicoddemus (changes requested):
Thanks @eyllanesc!

Other than the small comment, can you also please:

1. Add unittests for `_is_library_loaded` and `_get_backend_loaded`? 
2. Include a CHANGELOG entry.

eyllanesc:
In addition to the changelog change, this part of the README.rst should also be rewritten. 

https://github.com/pytest-dev/pytest-qt/blob/f00d7a484b915106ab74ff4f0e864d1a01c75341/README.rst#L79-L99

tests/test_basics.py
```diff
@@ -494,6 +498,73 @@ def _fake_import(name, *args):
...
+        "qDebug",
+        "qWarning",
+        "qCritical",
+        "qFatal",
+    ):
+        setattr(qtcore, method_name, lambda *_: None)
+
+    if backend in ("PyQt5", "PyQt6"):
+        pyqt_version = 0x050B00 if backend == "PyQt5" else 0x060000
+        setattr(qtcore, "PYQT_VERSION", pyqt_version + 1)
```
nicoddemus, line 529: Any reason why you are using `setattr` instead of just setting the attribute, like:

```suggestion
        qtcore.PYQT_VERSION = pyqt_version + 1
```

nicoddemus (approved):

nicoddemus:
Thanks @eyllanesc, and sorry for the delay!

PR #419, opened 2022-04-28
Drop Python 3.6, test on 3.11

The-Compiler:
Needed to get the CI green again.

### DIFF
--- a/tests/test_modeltest.py
+++ b/tests/test_modeltest.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 
 from pytestqt.qt_compat import qt_api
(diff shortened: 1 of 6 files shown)


tests/test_modeltest.py
```diff
@@ -94,12 +96,22 @@ def data(
     check_model(BrokenTypeModel(), should_pass=False)
 
 
+xfail_py311_pyside2 = pytest.mark.xfail(
+    sys.version_info[:2] == (3, 11) and qt_api.pytest_qt_api == "pyside2",
+    reason="Fails to OR mask flags",
+)
```
The-Compiler, line 102: Not sure if this is a PySide2 or Python 3.11 bug:

```pytb
>>> int(QtCore.Qt.AlignmentFlag.AlignHorizontal_Mask | QtCore.Qt.AlignmentFlag.AlignVertical_Mask)
TypeError: 'PySide2.QtCore.Qt.AlignmentFlag' object cannot be interpreted as an integer

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
SystemError: <class 'int'> returned a result with an exception set
```

bnavigator: This `XPASS`es for me on openSUSE Python 3.11.3 and PySide2 5.15.10

The-Compiler: @bnavigator Thanks, #497 should help.

PR #436, opened 2022-06-23
Update readme and changelog

The-Compiler:
Some small stuff we should probably fix before the release. Will self-merge once I get to the release if you didn't have a chance to look at it until then, as it seems trivial enough.


nicoddemus (approved):

nicoddemus:
Sure, LGTM!

PR #445, opened 2022-07-28
Update example github workflow in troubleshooting section of documentation

Czaki:
In this PR, I would like to update the GitHub example in the troubleshooting section to use more recent versions of workflows and python.


nicoddemus (approved):
Looks good, thanks!

PR #451, opened 2022-09-02
fix tiny docstring typo

k-dominik:
Heya,

not that this is super important or anything, and the warning also hints at the right method, but correct documentation is also not bad.

Cheers
D


nicoddemus (approved):
Thanks, appreciate it!

PR #460, opened 2022-10-19
Reference TimeoutError and ScreenshotError via qtbot

### LINKED ISSUE #459: Difference between `pytestqt.exceptions.TimeoutError` and Python's `TimeoutError`
Python has a `TimeoutError` already: https://docs.python.org/3/library/exceptions.html#TimeoutError

What is the purpose of `pytestqt.exceptions.TimeoutError`, which is not derived from `TimeoutError`? It is very easy write a

```python
try:
  ...
except  TimeoutError:
  pass
```

that does not work as intended because one is missing `from pytestqt.exceptions import TimeoutError` (which by the way gives:
> Redefining built-in 'TimeoutError'

in `pylint`).

I think `pytestqt.exceptions.TimeoutError` could be removed or, if it should be kept for compatibility, it should derive from `TimeoutError`.

nicoddemus:
Fix #459

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,23 @@
 repos:
+-   repo: https://github.com/PyCQA/autoflake
+    rev: v1.7.6
+    hooks:
+    -   id: autoflake
+        name: autoflake
+        args: ["--in-place", "--remove-unused-variables", "--remove-all-unused-imports"]
+        language: python
+        files: \.py$
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.3.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: debug-statements
+-   repo: https://github.com/asottile/pyupgrade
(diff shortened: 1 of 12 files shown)


nicoddemus:
cc @bersbersbers

The-Compiler (changes requested):
Thanks! I think the initial doc snippet in `signals.rst` should be updated as well, and maybe references to those exceptions in our tests too?

The-Compiler:
Sorry, I dropped the ball on this a bit - is it intended that 629c2cd6b23a65d4f3742c7520b98d80c77659dd includes changes to the pre-commit config?

The-Compiler (approved):
LGTM other than what I mention about the pre-commit config.

nicoddemus:
Hmm good catch, should be separate, let me rebase.

The-Compiler:
Looks like the pre-commit changes are still there? :thinking:

nicoddemus:
Ahh they are, but they are in a separate commit. :+1: 

I don't think we need a separate PR for those changes, or do you prefer a separate PR?

The-Compiler (approved):

PR #461, opened 2022-10-20
CI workflow improvements

nicoddemus:
* Drop 'checks' job from CI: this job is redundant now that we use pre-commit.ci and readthedocs builds docs for PRs.
* Split workflows into 'deploy' and 'build' for easier management.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v3
       with:
-        python-version: "3.10"
+        python-version: "3.11"
 
     - name: Build package
       run: |
@@ -30,7 +30,7 @@ jobs:
         python -m build
 
     - name: Publish package to PyPI
-      uses: pypa/gh-action-pypi-publish@v1.5.0
+      uses: pypa/gh-action-pypi-publish@v1.5.1
       with:
         user: __token__
         password: ${{ secrets.pypi_token }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,36 @@
+name: deploy
+
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
+
+jobs:
+
+  deploy:
+    if: github.repository == 'pytest-dev/pytest-qt'
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        persist-credentials: false
+
+    - name: Set up Python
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+        python-version: "3.10"
+
+    - name: Build package
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+        python -m build
+
+    - name: Publish package to PyPI
+      uses: pypa/gh-action-pypi-publish@v1.5.0
```
The-Compiler, line 33:
```suggestion
      uses: pypa/gh-action-pypi-publish@v1.5.1
```

while you're at it?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        persist-credentials: false
+
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
```
The-Compiler, line 24:
```suggestion
        python-version: "3.11"
```

now that it was released maybe?

.github/workflows/main.yml
```diff
@@ -58,50 +58,3 @@ jobs:
     - name: Test with tox
```
The-Compiler, line 58: While you're at it, maybe change `3.11-dev` to `3.11` above (released now), and also re-check if pyside6 works on it nowadays?

.github/workflows/main.yml
```diff
@@ -58,50 +58,3 @@ jobs:
...
-      with:
-        python-version: "3.7"
-    - name: Install tox
-      run: |
-        python -m pip install --upgrade pip
-        pip install tox
-    - name: Linting
-      run: |
-        tox -e linting
-    - name: Docs
```
The-Compiler, line 79: Do we really have a check failing the PR status from RTD, if a PR breaks the docs? Scrolling through the commit status list of this PR, I can't seem to find any?

The-Compiler (changes requested):

.github/workflows/main.yml
nicoddemus: Good catch, I have re-synced the webhook from RTD and is now working again:

image

nicoddemus:
Seems like 3.11 is not available for mac-os yet, let's wait a few days to try this again as this is not critical.

The-Compiler:
According to https://github.com/actions/setup-python/issues/531 this was fixed a day later :sweat_smile: I triggered a rerun.

The-Compiler (approved):
Looks good now!

nicoddemus:
Thanks!

PR #462, opened 2022-10-25
Stop importing from py

The-Compiler:
py._code has been moved to _pytest._code with pytest 2.9.0: https://docs.pytest.org/en/latest/changelog.html#id395

We still imported from py._code without actually depending from py. While this happened to work, it now breaks due to pytest dropping py: https://github.com/pytest-dev/pytest/pull/10396

Resulting in:

    File ".../src/pytestqt/logging.py", line 5, in <module>
        from py._code.code import TerminalRepr, ReprFileLocation
    ModuleNotFoundError: No module named 'py._code'; 'py' is not a package

with the latest pytest main.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,12 @@
 UNRELEASED
 ----------
 
+- Import the ``code`` sub-package from the correct location rather than the
+  deprecated ``py`` package, restoring compatibility with pytest 7.2.0, where
+  ``py`` was dropped. Thanks `@The-Compiler`_ for the PR.
+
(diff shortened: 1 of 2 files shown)


nicoddemus (approved):

nicoddemus:
Thanks!

The-Compiler:
Thank you for the quick merge! I assume you're planning to cut a new release after the currently open PRs are in?

nicoddemus:
Yep!

vkbo:
Thanks for the quick fix! I just discovered this error now and was investigating. And thanks for making the release so fast.

PR #472, opened 2023-01-25
Correct and clarify parameter names for qtbot.mouse*

oscargus:
There was an inconsistency between `stateKey` and `modifier` in the docs (`stateKey` in the signature, `modifier` in the parameters). Also, `pos` in the signature and `position` in the parameters.

This makes is consistent and clarifies that `stateKey` is used in the pyside bindings (both 2 and 6). (`modifier` is used in the PyQT bindings, not sure which one should be considered the "default". Although PySide* is loaded first, the benefit with `modifier` is that it is used in `key*`, which is referred to.)

As a side note, it could be an option for pytest-qt to adapt the keyword argument name depending on the binding.

Edit: I also added the enum Namespace where missing.

### DIFF
--- a/src/pytestqt/qtbot.py
+++ b/src/pytestqt/qtbot.py
@@ -59,30 +59,30 @@ class QtBot:
 
     Below are methods used to simulate sending key events to widgets:
 
-    .. staticmethod:: keyClick (widget, key[, modifier=Qt.NoModifier[, delay=-1]])
-    .. staticmethod:: keyClicks (widget, key_sequence[, modifier=Qt.NoModifier[, delay=-1]])
-    .. staticmethod:: keyEvent (action, widget, key[, modifier=Qt.NoModifier[, delay=-1]])
-    .. staticmethod:: keyPress (widget, key[, modifier=Qt.NoModifier[, delay=-1]])
-    .. staticmethod:: keyRelease (widget, key[, modifier=Qt.NoModifier[, delay=-1]])
+    .. staticmethod:: keyClick (widget, key[, modifier=Qt.KeyboardModifier.NoModifier[, delay=-1]])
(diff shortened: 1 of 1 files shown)


src/pytestqt/qtbot.py
```diff
@@ -59,30 +59,30 @@ class QtBot:
...
             also pressed. Possible flags are:
 
-            * ``Qt.NoModifier``: No modifier key is pressed.
-            * ``Qt.ShiftModifier``: A Shift key on the keyboard is pressed.
-            * ``Qt.ControlModifier``: A Ctrl key on the keyboard is pressed.
-            * ``Qt.AltModifier``: An Alt key on the keyboard is pressed.
-            * ``Qt.MetaModifier``: A Meta key on the keyboard is pressed.
-            * ``Qt.KeypadModifier``: A keypad button is pressed.
-            * ``Qt.GroupSwitchModifier``: X11 only. A Mode_switch key on the keyboard is pressed.
+            * ``Qt.KeyboardModifier.NoModifier``: No modifier key is pressed.
```
nicoddemus, line 79: To clarify: is this change only for PySide 2?

oscargus: This change is only required for PyQt6:

```
In [9]: from PyQt6.QtCore import Qt

In [10]: Qt.NoModifier
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[10], line 1
----> 1 Qt.NoModifier

AttributeError: type object 'Qt' has no attribute 'NoModifier'

In [11]: Qt.KeyboardModifier.NoModifier
Out[11]: <KeyboardModifier.NoModifier: 0>
```
In all other binding both versions exist. See, e.g., https://www.pythonguis.com/faq/pyqt5-vs-pyqt6/

nicoddemus: Awesome, thanks!

PR #478, opened 2023-02-01
migrate to qtpy

### LINKED ISSUE #184: Integration with qtpy
The support of multiple PyQt / PySide backends is a recurrent problem for any project working with Qt in Python. In pytest-qt it is solved by providing a compatibility `_QtApi` interface in `qt_compat.py` (that is not in the public interface). At the same time there are packages that specialize in providing such a compatibility interface, namely [spyder-ide/qtpy](https://github.com/spyder-ide/qtpy) and [mottosso/Qt.py](https://github.com/mottosso/Qt.py). For project that use one of those packages (e.g. matplotlib is [considering it](https://github.com/matplotlib/matplotlib/issues/6118)), and wanting to test it with with pytest-qt it would mean that they would not, technically be testing the project in the same conditions (and it would also mean setting differently named environment variables twice `PYTEST_QT_API` and `QT_API` or `QT_PREFERRED_BINDING`.) 

I think the integration could be straightforward: e.g. to integrate with qtpy, the following should almost work,

At the end of [**pytestqt/qt_compat.py**](https://github.com/pytest-dev/pytest-qt/blob/master/pytestqt/qt_compat.py#L192)
```py
try:
    if 'QT_API' not in os.environ and 'PYTEST_QT_API' in os.environ:
        os.environ['QT_API'] = os.environ['PYTEST_QT_API']
    import qtpy
    qt_api = qtpy
    qt_api.pytest_qt_api = qtpy.API
except ImportError:
     # fallback to the built-in compatibility layer..
    qt_api = _QtApi()
```

And maybe one extra environment in TravisCI for testing..

What do you think?

### LINKED ISSUE #477: Use qtpy instead of custom compatibility code
They support type hints, I would suggest to migrate to qtpy, I can help if you want

_Originally posted by @nrbnlulu in https://github.com/pytest-dev/pytest-qt/issues/474#issuecomment-1411980934_

nrbnlulu:
closes #477


nrbnlulu:
@nicoddemus Can you tell me wich tests are redundant by now?

tests/test_basics.py
```diff
@@ -566,7 +566,7 @@ def _fake_import(name, *args):
     def _fake_is_library_loaded(name, *args):
         return False
 
-    monkeypatch.delenv("PYTEST_QT_API", raising=False)
```
nicoddemus, line 569: Note that we need to keep supporting `PYTEST_QT_API` for backward compatibility, so we should not change these tests.

nicoddemus:
> @nicoddemus Can you tell me which tests are redundant by now?

I think the only test that is checking the implementation details related to qt-compat is `test_already_loaded_backend` (so it can be dropped), however the others should be adapted because they test user-facing functionality.

Which Qt backend to use is part of the user configuration, so it should remain tested.

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = py{37,38,39,310}-{pyqt5,pyside2,pyside6,pyqt6}, linting
 [testenv]
 deps=
     pytest
+    qtpy
```
nicoddemus, line 7: This should not be needed because `qtpy` is already a dependency of `pytest-qt` itself (now that I noticed it, `pytest` above is not needed either).

nrbnlulu: IIRC tests were failing without it

nrbnlulu:
> > @nicoddemus Can you tell me which tests are redundant by now?
> 
> I think the only test that is checking the implementation details related to qt-compat is `test_already_loaded_backend` (so it can be dropped), however the others should be adapted because they test user-facing functionality.
> 
> Which Qt backend to use is part of the user configuration, so it should remain tested.

We can refer to qtpy docs instead no?

nicoddemus:
> We can refer to qtpy docs instead no?

I would prefer to avoid breaking backward compatibility... probably it is a simple matter of:

```
if "PYTEST_QT_API" in os.environ:
    os.environ["QT_API"] = os.environ["PYTEST_QT_API"]
``` 

Before importing `qtpy`?

tox.ini
nicoddemus: Weird, can you double check? `qtpy` is already a dependency for the package, we should not have to duplicate it here... :thinking:

nicoddemus: Ahh in case you were already running `tox`, try passing `-r` for it to create a fresh environment.

src/pytestqt/plugin.py
```diff
@@ -226,18 +227,4 @@ def pytest_configure(config):
 
     if config.getoption("qt_log") and config.getoption("capture") != "no":
         config.pluginmanager.register(QtLoggingPlugin(config), "_qt_logging")
-
-    qt_api.set_qt_api(config.getini("qt_api"))
-
-
-def pytest_report_header():
```
The-Compiler, line 233: This should definitely stay - it's very useful information for test runs, and it missing also currently breaks a lot of self tests.

nicoddemus:
Closing this for now, we can reopen in the future as needed. :+1:

PR #480, opened 2023-03-01
Support an existing asyncio loop such as qasync

lbt:
When running tests in an async development environment there may already be a running loop which is assumed to be Qt compatible.

This approach allows the loop to be detected and used rather than creating a new one.

This test uses async Django calls and qasync

```
async def test_mqtest_save_db(qtbot, sensors, mqtest, db_record):
    async with qtbot.waitSignal(mqtest.signal_db_saved) as waiter:
        mqtest.db.tests = "New Value"
        mqtest.signal_save_db.emit()
    rec = await ATestRecord.objects.aget(pk=db_record.id)
    assert rec.tests == "New Value"
 ```

### DIFF
--- a/src/pytestqt/qtbot.py
+++ b/src/pytestqt/qtbot.py
@@ -495,6 +495,44 @@ class QtBot:
             yield
         spy.assert_not_emitted()
 
+    class Timeout:
+        import time
+        def __init__(self, timeout):
+            self.timeout = timeout
+            self.start = self.time.time()
+
+        def timed_out(self):
+            elapsed = self.time.time() - self.start
+            elapsed_ms = elapsed * 1000
+            return elapsed_ms > self.timeout
+
+        def msg(self):
+            return f"waitUntil timed out in {self.timeout} milliseconds"
+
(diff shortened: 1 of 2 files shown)


lbt:
This "works for me" but I've not written tests etc until someone can confirm the approach is acceptable :)

nicoddemus:
Thanks @lbt for the PR!

To be honest however, I personally don't work with `asyncio`, so my knowledge is limited, which makes me hesitant to include this kind of support, given me and @The-Compiler will have to maintain that code going forward.

Having said that, this probably won't work with other async frameworks (such as `trio`, which I very much enjoy), which makes me wonder if changing the signal blockers directly is the way to go. Perhaps using a new class, which receives a `qtbot` in the constructor is the more extensible approach? In that case, it would not even need to be introduced into pytest-qt, it could be its own package.

But let's wait for @The-Compiler to chime in. 👍

lbt:
I've been extending this a little for my personal use case and seeing how other features need handling.
I refactored a_awaitUntil to be more DRY.

Issue #504, opened 2023-07-16
Segfault on Ubuntu 22.04


hakonhagland:
I am on Ubuntu 22.04, using [Poetry](https://python-poetry.org/), Python 3.10 and PyQt6. When I run a test with pytest using the `qtbot` fixture I get a segmentation fault. If I remove the `qtbot` fixture from the test file, it works fine. Here is a minimal example (my real test case is more complicated):

    $ poetry new --src myproject
    $ cd myproject
    $ poetry add pyqt6
    $ poetry add --group=dev pytest
    $ poetry add --group=dev pytest-qt
    $ cat <<'END' > pytest.ini
    [pytest]
    qt_api=pyqt6
    END
    $ cat <<'END' > src/myproject/main.py
    from PyQt6.QtWidgets import QApplication
    def qapp(app: QApplication) -> None:  # This is the function that we will test with pytest
        app.quit()
    END
    $ cat <<'END' > tests/test_main.py  # This is the test file
    import pytest
    from PyQt6.QtWidgets import QApplication
    import myproject.main
    from typing import Any
    QtBot = Any  # Missing type hints here
    def test_app(qtbot: QtBot) -> None:
        app = QApplication([])
        myproject.main.qapp(app)
        assert 1 == 1
    END
    $ poetry install
    $ poetry shell
    $ pytest
    ============================================================================ test session starts ============================================================================
    platform linux -- Python 3.10.4, pytest-7.4.0, pluggy-1.2.0
    PyQt6 6.5.1 -- Qt runtime 6.5.1 -- Qt compiled 6.5.1
    rootdir: /home/hakon/test/python/pytest-qt/qtbot/myproject
    plugins: qt-4.2.0
    collected 1 item                                                                                                                                                            
    
    tests/test_main.py Fatal Python error: Segmentation fault
    
    Current thread 0x00007f24e2fc1740 (most recent call first):
      File "/home/hakon/.cache/pypoetry/virtualenvs/myproject-rfHFWcwh-py3.10/lib/python3.10/site-packages/_pytest/python.py", line 194 in pytest_pyfunc_call
      File "/home/hakon/.cache/pypoetry/virtualenvs/myproject-rfHFWcwh-py3.10/lib/python3.10/site-packages/pluggy/_callers.py", line 80 in _multicall
[...]

PR #505, opened 2023-07-16
Add more info about qApp pointer

hakonhagland:
Add more info in the documentation about how to get the qApp pointer when using the qtbot fixture, see issue #504 for more information.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -33,9 +33,18 @@ To test this widget's basic functionality, create a test function:
         tmp_path.joinpath("video2.srt").touch()
 
 Here the first parameter indicates that we will be using a ``qtbot`` fixture to control our widget.
+
+.. note::
+
+    It is not necessary to create a QApplication instance, since the ``qtbot`` fixture will
+    do this for you. The ``QApplication`` object is accessible through the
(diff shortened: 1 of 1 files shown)


nicoddemus (approved):
Thanks!

hakonhagland:
> Add more info in the documentation about how to get the qApp pointer

@nicoddemus Sorry, I did not check the sphinx output. I think there is something wrong with the backticks:

image

I will try to fix it.

hakonhagland:
> The `QApplication` object is accessible through the  `QApplication.instance()` method

I also discovered that you can use the `qapp_cls` fixture

https://github.com/pytest-dev/pytest-qt/blob/f1a6ac554c1d2a81392ed526c56c49265b6a02ea/src/pytestqt/plugin.py#L32

to get the object. Maybe we should also mention this alternative in the documentation?

hakonhagland:
> qapp will give you the QApplication instance.

@The-Compiler Thanks for the information! I have updated PR #506 such that the docs mentions this `qapp` fixture instead of `QApplication.instance()`

nicoddemus:
> @nicoddemus Sorry, I did not check the sphinx output. I think there is something wrong with the backticks:

No worries, I missed that as well. Thanks for the follow up!

PR #508, opened 2023-07-19
qtbot.click() does not exist?

hakonhagland:
As far as I can tell there is no function called `click()` in the `QtBot` class.

Removes the mention of `qtbot.click()` from the documentation, and replaces it with `qtbot.mouseClick()` instead.


nicoddemus (approved):
Thanks, appreciate it!

Issue #514, opened 2023-08-28
How to reset qapp_args for each test class?


hakonhagland:
I would like to run each test class with a different set of command line arguments (`qapp_args` fixture), here is a minimal example:

```python
import sys
import pytest
from PyQt6.QtCore import QCommandLineParser
from PyQt6.QtWidgets import QApplication

def get_positional_args(app):
    app.setApplicationName("MyApp")
    app.setApplicationVersion("0.1")
    parser = QCommandLineParser()
    parser.addHelpOption()
    parser.addVersionOption()
    parser.addPositionalArgument("my_pos_arg1", "description of my_pos_arg1")
    parser.addPositionalArgument("my_pos_arg2", "description of my_pos_arg2")
    parser.process(app)
    arguments = parser.positionalArguments()
    return arguments

if __name__ == '__main__':
    app = QApplication(sys.argv)
    args = get_positional_args(app)
    print(args)

class Test1:
    @pytest.fixture(scope="session")
    def qapp_args(self):
        return ["prog_name", "a1"]

    def test_pos_args(self, qapp):
        args = get_positional_args(qapp)
        assert len(args) == 1

class Test2:
    @pytest.fixture(scope="session")
    def qapp_args(self):
        return ["prog_name", "a1", "a2"]

    def test_pos_args(self, qapp):
        args = get_positional_args(qapp)
        assert len(args) == 2
```

But in the second test class `Test2` overriding `qapp_args()` does not work. It seems like it is using the `qapp_args()` defined in the `Test1` class instead. How can I solve this?

![image](https://github.com/pytest-dev/pytest-qt/assets/6708379/eaf65b00-86c7-47e6-8edc-4d5c86840562)

hakonhagland:
> It seems like it is using the qapp_args() defined in the Test1 class instead.

I think the fixtures should have scope `class` and not `session`? But if I change the scope to `class` I get another error:
`ScopeMismatch: You tried to access the class scoped fixture qapp_args with a session scoped request object`:

![image](https://github.com/pytest-dev/pytest-qt/assets/6708379/d45ecf5c-de76-446f-880d-b761e1fa9bc6)

hakonhagland:
> ScopeMismatch: You tried to access the class scoped fixture qapp_args with a session scoped request object

I think it might not be possible to change the `qapp` fixture per test class, since the fixture creates a `QApplication` instance

[...]

PR #515, opened 2023-08-28
Pass program name to QApplication contructor

### LINKED ISSUE #483: QtWebEngine 6.5: Fatal Python error: Aborted / Argument list is empty, the program name is not passed to QCoreApplication. base::CommandLine cannot be properly initialized.
With the upcoming QtWebEngine 6.5, a snippet like:

```python
from PyQt6.QtWidgets import QApplication
from PyQt6.QtWebEngineCore import QWebEngineProfile

app = QApplication([])
QWebEngineProfile()
```

fails with:

```
Argument list is empty, the program name is not passed to QCoreApplication. base::CommandLine cannot be properly initialized.
```

due to:

- [[QTBUG-110157] Incorrect command-line parsing if argc=0 - Qt Bug Tracker](https://bugreports.qt.io/browse/QTBUG-110157)
- [Improve error handling of argument parsing (Ibf14b11b) · Gerrit Code Review](https://codereview.qt-project.org/c/qt/qtwebengine/+/455347)
- [Yield fatal error for empty command-line arguments (If0144af8) · Gerrit Code Review](https://codereview.qt-project.org/c/qt/qtwebengine/+/457045)

Which, with pytest:

```python
import pytest

from PyQt6.QtWebEngineCore import QWebEngineProfile
from PyQt6.QtWebEngineWidgets import QWebEngineView

def test_qtwe(qapp):
    print(f"args: {qapp.arguments()}")
    QWebEngineProfile.defaultProfile()
```

just turns into:

```
test_qtwe.py Fatal Python error: Aborted
```

and a stacktrace, due to output capturing...

This is because we also default to empty args:

https://github.com/pytest-dev/pytest-qt/blob/f0dced8f7918d6a19353c2d99cc1e77fd948b6c0/src/pytestqt/plugin.py#L14-L28

As a workaround, this works:

```python
import sys
import pytest

from PyQt6.QtWebEngineCore import QWebEngineProfile
from PyQt6.QtWebEngineWidgets import QWebEngineView

@pytest.fixture(scope="session")
def qapp_args():
    return [sys.argv[0]]

def test_qtwe(qapp):
    print(f"args: {qapp.arguments()}")
    QWebEngineProfile.defaultProfile()
```

but I expect that many people run into this when Qt 6.5 comes out, [currently planned](https://wiki.qt.io/Qt_6.5_Release) for the 30th.

In addition to that, [our documentation](https://pytest-qt.readthedocs.io/en/latest/reference.html#pytestqt.plugin.qapp_args) is just wrong too. It suggests:

```python
@pytest.fixture(scope="session")
def qapp_args():
    return ["--arg"]
```

but that would mean that `--arg` is now `sys.argv[0]` (the script name), which makes no sense at all...

---

So how do we fix this? One proposal would be to always prepend `sys.argv[0]` to the given `qapp_args`. This would mean that the example is now correct, and existing testsuites would continue working with QtWebEngine 6.5 with no further changes required.

However, it's a backwards-**in**compatible change, and if someone already accounts for this in a `qapp_args`, the `argv[0]` would now be passed *twice*...

A [cursory GitHub search](https://github.com/search?q=%22def+qapp_args%22+-path%3Asrc%2Fpytestqt%2Fplugin.py+-path%3Apytestqt%2Fplugin.py+-path%3Atests%2Ftest_basics.py&type=code) reveals that projects already do this (lots of qutebrowser forks there :laughing:):

- [eddy/conftest.py at 2b257070ca15d63e31c8f53f0c538cbcddee3e7a · obdasystems/eddy](https://github.com/obdasystems/eddy/blob/2b257070ca15d63e31c8f53f0c538cbcddee3e7a/tests/conftest.py#L72)
- [COUNTER-5-Report-Tool/test_ImportFile.py at 5d70a8a8483cb0e1718fb22dbe025437d5076695 · CS-4820-Library-Project/COUNTER-5-Report-Tool](https://github.com/CS-4820-Library-Project/COUNTER-5-Report-Tool/blob/5d70a8a8483cb0e1718fb22dbe025437d5076695/tests/test_ImportFile.py#L18)

The second option would be to fix our doc examples, *and* fix the default `qapp_args` to `return [sys.argv[0]]` instead. This is technically still a backwards-incompatible change, but also clearly a bugfix for something we are doing wrong and never noticed... However, it means that projects with a custom `qapp_args` (I could only find qutebrowser) will need to adjust.

---

@nicoddemus what do you think? Do you have the capacity to fix this and cut a new release in the coming few days? Happy to do reviews and test with Qt 6.5 locally, but I'm a bit busy giving a company training between the 20th and 30th... and by the time I'm back, Qt 6.5 will already be released most likely.

hakonhagland:
`QCommandLineParser` needs `QApplication` to be initialized with a list with at least one argument: The name of the program.

See https://github.com/pytest-dev/pytest-qt/issues/514#issuecomment-1696462533 for an example.

Closes #483.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,5 @@
-4.3.0 (2023-09-20)
-------------------
+UNRELEASED
+-----------
 
 - ``qapp`` now sets up the ``QApplication`` instance with a command line argument like this
   ``QApplication([prog_name])`` instead of using an empty list ``QApplication([])``.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,20 @@
+4.3.0 (2023-09-20)
+------------------
+
+- ``qapp`` now sets up the ``QApplication`` instance with a command line argument like this
+  ``QApplication([prog_name])`` instead of using an empty list ``QApplication([])``.
+  Here ``prog_name`` is the name of the app which defaults to ``pytest-qt-app``, but can
+  be redefined in the ``pytest.ini`` file, see :ref:`qapp fixture<setting-qapp-name>`.
+  Alternatively, the arguments that will be passed to ``QApplication`` can be defined
+  explicitly using the ``qapp_args`` fixture. This means that the default behavior of
+  the ``qapp_args`` fixture is now also changed accordingly: it now returns the list
+  ``[prog_name]`` instead of an empty list. Thanks to `@The-Compiler`_ (`#483`_) and
+  `@hakonhagland`_ (`#515`_).
+
+..  _#515: https://github.com/pytest-dev/pytest-qt/pull/515
+..  _#483: https://github.com/pytest-dev/pytest-qt/issues/483
+
+
 4.2.0 (2022-10-25)
 ------------------
 
@@ -731,6 +748,7 @@ First working version.
 .. _@fabioz: https://github.com/fabioz
 .. _@fogo: https://github.com/fogo
 .. _@gqmelo: https://github.com/gqmelo
+.. _@hakonhagland: https://github.com/hakonhagland
 .. _@itghisi: https://github.com/itghisi
 .. _@jdreaver: https://github.com/jdreaver
 .. _@mitya57: https://github.com/mitya57
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -546,13 +546,35 @@ def test_qapp_args(testdir):
 
         @pytest.fixture(scope='session')
         def qapp_args():
-            return ['--test-arg']
+            return ['prog_name', '--test-arg']
         """
     )
     testdir.makepyfile(
         """
         def test_args(qapp):
             assert '--test-arg' in list(qapp.arguments())
+    """
+    )
+    result = testdir.runpytest_subprocess()
+    result.stdout.fnmatch_lines(["*= 1 passed in *"])
+
+
+def test_qapp_args_default(testdir):
+    """
+    Test QApplication default arguments.
+    """
+
+    #    testdir.makeini(
+    #        """
+    #        [pytest]
+    #        qt_qapp_name = frobnicator
+    #        """
+    #    )
+    testdir.makepyfile(
+        """
+        def test_args(qapp):
+            args = qapp.arguments()
+            assert args[0] == 'pytest-qt-qapp'
     """
     )
     result = testdir.runpytest_subprocess()
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -564,12 +564,6 @@ def test_qapp_args_default(testdir):
     Test QApplication default arguments.
     """
 
-    #    testdir.makeini(
-    #        """
-    #        [pytest]
-    #        qt_qapp_name = frobnicator
-    #        """
-    #    )
     testdir.makepyfile(
         """
         def test_args(qapp):
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,6 +43,6 @@ repos:
     -   id: rst
         name: rst
         entry: rst-lint --encoding utf-8
-        files: ^(CHANGELOG.rst|HOWTORELEASE.rst|README.rst)$
+        files: ^(HOWTORELEASE.rst|README.rst)$
         language: python
         additional_dependencies: [pygments, restructuredtext_lint]
--- a/docs/qapplication.rst
+++ b/docs/qapplication.rst
@@ -75,6 +75,8 @@ If your tests require access to app-level functions, like
 The ``qapp`` fixture will then use the returned class instead of the default
 ``QApplication`` from ``QtWidgets``.
 
+.. _setting-qapp-name:
+
 Setting a QApplication name
 ---------------------------
 
--- a/src/pytestqt/plugin.py
+++ b/src/pytestqt/plugin.py
@@ -12,7 +12,7 @@ from pytestqt.qtbot import QtBot, _close_widgets
 
 
 @pytest.fixture(scope="session")
-def qapp_args():
+def qapp_args(pytestconfig):
     """
     Fixture that provides QApplication arguments to use.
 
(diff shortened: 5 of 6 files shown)


nicoddemus:
Is this related to #483 somehow?

The-Compiler:
> Is this related to #483 somehow?

Looks like it is, and should probably close that issue? This definitely needs a changelog entry and a test as well though.

hakonhagland:
> this should be sys.argv[0] instead. The QApplication.applicationName() isn't necessarily the same thing.

Right, but `sys.argv[0]` will be something like ` /home/hakon/.pyenv/versions/3.10.4/bin/pytest` if the test is run with `pytest`, right?

The-Compiler:
True, but I'd argue that is the correct value. The whole point of `sys.argv[0]` is for your code to know via which command it was invoked, and I don't think we should mask that. No strong feelings on it though, I can see an argument for either side now that you mention it.

hakonhagland:
> I can see an argument for either side now that you mention it.
 
@The-Compiler OK fine! I'll then keep the original version for now, if it is ok?

> This definitely needs a changelog entry and a test as well though.

Added a unit test.

hakonhagland:
> This definitely needs a changelog entry 

@The-Compiler Should I add an entry in the changelog in this PR, or should it be a separate PR?

hakonhagland:
@nicoddemus In `CHANGELOG.rst` can I use for example `` :ref:`qapp fixture<setting-qapp-name>` `` to cross reference a section in the documentation? It does not seem to work, I get `ERROR CHANGELOG.rst:4 Unknown interpreted text role "ref"` when compiling.

hakonhagland:
> It does not seem to work

I think it does work, but the linter 

https://github.com/pytest-dev/pytest-qt/blob/da33817f6f70cf3440cc1e3155c814136bd68dcd/.pre-commit-config.yaml#L45

does not see the whole picture?

hakonhagland:
> but the linter does not see the whole picture?

Is it possible to make the linter somehow ignore this line with `:ref:` ?

nicoddemus:
@hakonhagland hmm I don't think it is possible to configure the linter that way, perhaps we can exclude `CHANGELOG.rst` from the linter instead?

```
files: ^(CHANGELOG.rst|HOWTORELEASE.rst|README.rst)$
```

Let's remove `CHANGELOG.rst` from there, the changelog file is included into the docs, so Sphinx will check that for us already.

hakonhagland:
> Let's remove CHANGELOG.rst from there

Ok done! Please review last commit for changelog update

CHANGELOG.rst
```diff
@@ -1,3 +1,20 @@
+4.3.0 (2023-09-20)
+------------------
```
nicoddemus, line 2: We probably won't have an immediate release, but even so the process is to set version/date at the time of release. 👍 

```suggestion
UNRELEASED
----------
```

tests/test_basics.py
```diff
@@ -559,6 +559,28 @@ def test_args(qapp):
...
+    """
+    Test QApplication default arguments.
+    """
+
+    #    testdir.makeini(
+    #        """
+    #        [pytest]
+    #        qt_qapp_name = frobnicator
+    #        """
+    #    )
```
nicoddemus, line 572:
```suggestion
```

nicoddemus:
> @The-Compiler OK fine! I'll then keep the original version for now, if it is ok?

Do you have a strong opinion on this @hakonhagland?

hakonhagland:
> Do you have a strong opinion on this

Do you mean whether to use `sys.argv[0]` or `qt_qapp_name` from `pytest.ini` as the default value for `qapp_args` ?

nicoddemus:
> Do you mean whether to use sys.argv[0] or qt_qapp_name from pytest.ini as the default value for qapp_args ?

Yes, I'm kind leaning towards @The-Compiler 's original opinion of using `sys.argv[0]`: seems correct to use `/path/to/pytest` in that case given we are running the tests, not the actual application.

But only if you do not have a strong opinion on this.

hakonhagland:
> seems correct to use /path/to/pytest

I think the QApplication should not depend on the location of the `pytest` script in any way. So we should not pass that to  QApplication constructor. But the test scripts `test_*.py` might depend on that, but I have not come up with a good example of this yet. If a test script needs to know the location of  the `pytest` script it should be able to recover that  by accessing `sys.argv[0]` directly right?

nicoddemus:
> I think the QApplication should not depend on the location of the pytest script in any way. 

Hmm not sure what you mean, `QApplication` itself does not care what the first argument is (it just cares if there are no arguments).

> But the test scripts test_*.py might depend on that, but I have not come up with a good example of this yet. If a test script needs to know the location of the pytest script it should be able to recover that by accessing sys.argv[0] directly right?

Not sure about what what some tests might *need* hypothetically, can't think of a situation where someone wants to check the first argument passed to `QApplication`.

Another point is that the the recommendation in production code is to pass `sys.argv`, not an "application name + arguments".

hakonhagland:
> the recommendation in production code is to pass sys.argv, not an "application name + arguments".

@nicoddemus I think this is for code running normally, not for code being executed by another program like pytest. Therefore  it will be safe and also the correct approach to pass application name instead of the location of the pytest script (that is: sys.argv[0])

nicoddemus:
> @nicoddemus I think this is for code running normally, not for code being executed by another program like pytest. Therefore it will be safe and also the correct approach to pass application name instead of the location of the pytest script (that is: sys.argv[0])

I'm not entirely convinced but TBH I'm not sure this will really matter much in the end, so I'm fine with we going in that direction. 👍 

Letting @The-Compiler do another round of review/merge. 👍 

Thanks again @hakonhagland for the contribution!

nicoddemus (approved):

The-Compiler (dismissed):
Looks good all in all! Not sure what's up with the failing CI environments, and no time right now to dig into that.

nicoddemus:
> Looks good all in all! Not sure what's up with the failing CI environments, and no time right now to dig into that.

They are all related to pyside6 and unrelated to this PR, except for the pre-commit job:

```
tests/test_wait_signal.py:620:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```

Just pushed a fix to that one.

nicoddemus:
Thanks @hakonhagland!

PR #516, opened 2023-08-28
Set sphinx module path

hakonhagland:
Sphinx module path should be set to `../src` not `..` for sphinx autodoc extension to work locally when typing `make html` from the `docs` folder. If not, `sphinx-build` might use a source file from pytest-qt installed in the python `site-packages` directory outside the git repo which is obviously not correct (and very frustrating to debug :) )


nicoddemus:
Thanks! Seems we never noticed because on RTD we install pytest-qt on the virtualenv. 👍

PR #525, opened 2023-10-10
Add note about PySide6 exception capture and fix tests

nicoddemus:
PySide6 does its own exception capture during events, re-raising them when control gets back to Python.

This feature causes our own exception capturing mechanism to not work, but it is not needed anyway, because the exception re-raising done by PySide6 contains a better traceback than it is possible for pytest-qt to provide, given it can re-raise at the exact point where control was given back to Python.

Besides noting this in the documentation, also accomodate the tests.


The-Compiler:
I don't know much about PySide, and haven't looked into this in detail - but shouldn't this be enabled only for PySide 6.5.2+ and 6.6+ rather than all PySide6 versions? A look at the CI reveals that [6.5.1 passed](https://github.com/pytest-dev/pytest-qt/actions/runs/5603903181/job/15181562522#step:5:24) and [6.5.2 started to fail](https://github.com/pytest-dev/pytest-qt/actions/runs/6010782935/job/16302918036#step:5:24), and some digging in the [Qt for Python Development Notes](https://wiki.qt.io/Qt_for_Python_Development_Notes#8._June_2023) revealed [PYSIDE-2335](https://bugreports.qt.io/browse/PYSIDE-2335) which indeed was merged in 6.5.2 and 6.6.

nicoddemus:
> I don't know much about PySide, and haven't looked into this in detail

Oh I should have been clear, I also did not look into detail, this was just the behavior I've observed today after attempting to fix the tests -- you did a more through dig than I did. 😁 

> but shouldn't this be enabled only for PySide 6.5.2+ and 6.6+ rather than all PySide6 versions?

What do you mean by "enabled"? Note that I did not change any production code, only accommodated the tests. This was not really a change in pytest-qt, just a realization that the behavior has changed, and we updated the tests accordingly. I do not think we should support different pyside6 versions in the test suites, if that's what you mean: we do not have the bandwidth for this kind of "fine" support, a test suite which works in PySide6 <6.5.2 and PySide6 <=6.5.2.

Or do you mean to mention 6.5.2+ in the documentation? That is fair.

nicoddemus:
@The-Compiler gentle ping.

The-Compiler:
Whoops, sorry for the delay! I have quite a backlog of stuff after ignoring my mails for a while because I moved...

I thought we already tested with different Qt versions, but indeed we don't (it's something I do for qutebrowser). So I agree it's fine to leave this as-is. Would indeed be good to clarify this in the docs though, to avoid confusing people with that statement.

nicoddemus:
Thanks!

PR #542, opened 2024-02-07
Use new-style hook wrappers

nicoddemus:
The [new-style hook wrappers](https://pluggy.readthedocs.io/en/stable/#wrappers) are easier to use because you do not need to call `outcome.force_exception` on errors, and doing so prevents other hook callers from running.

This requires pluggy>=1.1.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 UNRELEASED
 ----------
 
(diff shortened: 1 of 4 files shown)


nicoddemus:
Going ahead and merge this as it is just a small internal improvement. 👍

PR #572, opened 2024-10-08
Drop Python 3.8 support, add 3.13

nicoddemus:


nicoddemus:
Wait a bit until actions/setup-python publishes official 3.13 packages.

nicoddemus:
Thanks for the ping, indeed I had completely forgotten about this one.

Rebased, let's see.

nicoddemus:
I don't think PySide2 is still supported, the last release was in 2022, and of course there is no more releases for more recent Python versions.

Perhaps we should drop official support for it as well?

The-Compiler (dismissed):
Thanks for picking things up again, looks great!

The-Compiler:
> I don't think PySide2 is still supported, the last release was in 2022, and of course there is no more releases for more recent Python versions.
> 
> Perhaps we should drop official support for it as well?

I suppose we could, though not sure how much there is to gain if we still keep Qt5 (via PyQt5) around? https://pypi.org/project/PyQt5/#history seems alive enough to keep it IMHO.

nicoddemus:
I personally have an interest in PyQt5 because that's what we still use at ESSS, and is still maintained.

I will open a PR removing pyside2 and we can evaluate if it is worth or not.

PR #581, opened 2024-12-10
ci: Switch to download-artifact v4

The-Compiler:
See https://github.com/actions/upload-artifact?tab=readme-ov-file#v4---whats-new from what I can tell, we're not affected by the breaking changes.

The v3 action will stop working on January 30th:
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


The-Compiler:
Superseded by #582.

PR #583, opened 2024-12-11
Drop support for PySide2

nicoddemus:
PySide2 is no longer maintained, with the last release being made in 2022.

Follow-up to https://github.com/pytest-dev/pytest-qt/pull/572#issuecomment-2535835704.

@The-Compiler I like the removed cruft and CI jobs, but I understand if you would like to keep the support around.

### DIFF
--- a/tests/test_wait_signal.py
+++ b/tests/test_wait_signal.py
@@ -881,7 +881,7 @@ class TestAllArgs:
 def get_mixed_signals_with_guaranteed_name(signaller):
     """
     Returns a list of signals with the guarantee that the signals have names (i.e. the names are
-    manually provided in case of using PySide2, where the signal names cannot be determined at run-time).
+    manually provided in case of using PySide6, where the signal names cannot be determined at run-time).
     """
     if qt_api.is_pyside:
         signals = [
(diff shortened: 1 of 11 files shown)


tests/test_wait_signal.py
```diff
@@ -913,27 +913,6 @@ def test_empty_when_no_signal(self, qtbot, signaller):
             pass
         assert blocker.all_signals_and_args == []
 
-    def test_empty_when_no_signal_name_available(self, qtbot, signaller):
```
The-Compiler, line 916: I'm a bit confused by the current situation here (also for the test below) - is this really only relevant for PySide2, or did we just never add PySide6 to the test?

From what this looks like:

https://github.com/pytest-dev/pytest-qt/blob/89178edf898b635a21c59342a0771df55170f160/src/pytestqt/wait_signal.py#L104-L123

and from [[PYSIDE-1911] Provide a way to get the name of a Signal(Instance) - Qt Bug Tracker](https://bugreports.qt.io/browse/PYSIDE-1911) only being partially fixed (I think we never implemented the "extract from repr" part?), I believe those tests should still be run on PySide6 actually?

nicoddemus: Ahh great catch! You are right, they should be running for PySide 6 too. 

Brought the tests back and updated the condition.

The-Compiler (approved):

Issue #232, opened 2018-09-01
Proper tests for qInfo when it becomes fully available


nicoddemus:
Followup to #225, `qInfo` is not currently exposed by the bindings. As suggested in https://github.com/pytest-dev/pytest-qt/issues/225#issuecomment-417868188, we can workaround that by using `QMessageLogger`, but sadly that's not exposed by `PySide` yet.

For now we can make a simple test for INFO messages, but when `QMessageLogger` is available (or even better `qInfo`) we should update the other tests to issue/capture INFO messages.

The-Compiler:
Closing this as the PR was merged. I also [sent a mail](https://www.riverbankcomputing.com/pipermail/pyqt/2018-September/040862.html) to the PyQt mailinglist, so I'd expect it to show up in PyQt soon :wink:

nicoddemus:
Thanks!

The-Compiler:
FWIW I got the usual "Should be in tonight's snapshot." answer, so `qInfo()` should be in PyQt 5.11.3, which is planned for after Qt 5.11.2, which in turn is planned the coming week or so :wink:

PR #593, opened 2025-03-11
Enable qInfo tests for PySide6

penguinpee:
PySide6 now exposes `qInfo` just like Qt6. This enables the `qInfo`
tests accordingly.

For reference see issue #232.


nicoddemus (approved):
Awesome, thanks! Appreciate it!

The-Compiler:
This breaks compatibility with PySide6 < 6.8.2. Reverted and pushed an alternative fix in #597.

Issue #586, opened 2024-12-24
AttributeError: 'NoneType' object has no attribute 'start'


mmerlo:
I'm seeing an issue with pytest-qt (4.4.0) and I am not sure if it is on my end or an issue with pytest-qt. Most of the time, the test will pass, but this issue will cause the test to fail about 10% to 20% of the time. Below is code that reproduces the issue:
```
from PySide6.QtCore import QThread, QObject, Signal
import traceback

class MyRelay(QObject):
    sig_to_relay = Signal()

class MyWorker(QObject):
    my_signal = Signal()

    def __init__(self, relay):
        super().__init__()
        self._relay = relay
        self._relay.setParent(self)
        self._relay.sig_to_relay.connect(self.my_signal)

    
class MyThread(QObject):
    def __init__(self):
        self._workers = None
        self._thread = QThread()

    def add_worker(self, worker):
        self._workers  = worker

    def start(self):
        self._workers.moveToThread(self._thread)
        self._thread.start()


class TestThread:
    def test_error_1(self, qtbot):
        relay = MyRelay()
        mw = MyWorker(relay)
        my_thread = MyThread()
        my_thread.add_worker(mw)
        my_thread.start()

        try:
            with qtbot.waitSignal(
                mw.my_signal, timeout=500) as blocker:
                relay.sig_to_relay.emit()
                print(blocker)
        except Exception as e:
            traceback.print_exc()
            raise e
        my_thread._thread.quit()
        my_thread._thread.wait()
```
In this example, the `try` block is to catch the exception before it fails the test so that I can check the traceback. During failures, the traceback is:

```
Traceback (most recent call last):
  File "~python_work/pytest_qt_qthread/test_pytest_qt.py", line 41, in test_warning_message
    with qtbot.waitSignal(
  File "~python_work/pytest_qt_qthread/.venv/lib/python3.12/site-packages/pytestqt/wait_signal.py", line 151, in __exit__
    self.wait()
  File "~python_work/pytest_qt_qthread/.venv/lib/python3.12/site-packages/pytestqt/wait_signal.py", line 48, in wait
    self._timer.start()
    ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'start'
```
It appears that the a timer in pytest-qt is getting through the `timer is None` check but then is `None` when `self._timer.start()` is called a few lines later.

[...]

Issue #558, opened 2024-05-28
SystemError on disconnect with pyside 6.7.1


tlambert03:
Don't see an existing issue for this yet, so opening one.  This seems related to https://github.com/pytest-dev/pytest-qt/issues/552, but has now turned into an exception with pyside 6.7.1:

```python
from PySide6.QtWidgets import QLineEdit

def test_thing(qtbot):
    wdg = QLineEdit()
    qtbot.addWidget(wdg)
    with qtbot.waitSignal(wdg.textChanged):
        wdg.setText("hello")
```

running this test raises:

```
tests/test_thing.py F                                                    [100%]

=================================== FAILURES ===================================
__________________________________ test_thing __________________________________
CALL ERROR: Exceptions caught in Qt event loop:
________________________________________________________________________________
RuntimeWarning: Failed to disconnect (<bound method _AbstractSignalBlocker._quit_loop_by_timeout of <pytestqt.wait_signal.SignalBlocker object at 0x113115c50>>) from signal "timeout()".

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/talley/miniforge3/envs/superqt/lib/python3.11/site-packages/pytestqt/wait_signal.py", line 219, in _quit_loop_by_signal
    self._cleanup()
  File "/Users/talley/miniforge3/envs/superqt/lib/python3.11/site-packages/pytestqt/wait_signal.py", line 224, in _cleanup
    super()._cleanup()
  File "/Users/talley/miniforge3/envs/superqt/lib/python3.11/site-packages/pytestqt/wait_signal.py", line 66, in _cleanup
    _silent_disconnect(self._timer.timeout, self._quit_loop_by_timeout)
  File "/Users/talley/miniforge3/envs/superqt/lib/python3.11/site-packages/pytestqt/wait_signal.py", line 741, in _silent_disconnect
    signal.disconnect(slot)
SystemError: <method 'disconnect' of 'PySide6.QtCore.SignalInstance' objects> returned a result with an exception set
________________________________________________________________________________
----------------------------- Captured stderr call -----------------------------
Exceptions caught in Qt event loop:
________________________________________________________________________________
RuntimeWarning: Failed to disconnect (<bound method _AbstractSignalBlocker._quit_loop_by_timeout of <pytestqt.wait_signal.SignalBlocker object at 0x113115c50>>) from signal "timeout()".

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

Issue #552, opened 2024-04-03
"RuntimeError: Failed to disconnect" with `PySide6==6.7.0`


bersbersbers:
I recently tried upgrading to an early `PySide6==6.7.0` (via https://download.qt.io/snapshots/ci/pyside/6.7/latest/pyside6), and I find that `pytest` then issues a lot of warnings that I cannot seem to suppress:

>   c:\Code\project\.venv\Lib\site-packages\pytestqt\wait_signal.py:741: RuntimeError: Failed to disconnect (<bound method _AbstractSignalBlocker._quit_loop_by_timeout of <pytestqt.wait_signal.SignalBlocker object at 0x000001BE3BDA6390>>) from signal "timeout()".
    signal.disconnect(slot)

`filterwarnings = ["ignore"]` does not seem to have an effect, possibly because `RuntimeError` is not a proper warning.

nicoddemus:
It is probably because the exception is being raised inside the Qt event loop.

penguinpee:
I'm seeing the same here:

```python
=================================== FAILURES ===================================
________________________________ test_destroyed ________________________________
CALL ERROR: Exceptions caught in Qt event loop:
________________________________________________________________________________
RuntimeError: Internal C++ object (Obj) already deleted.
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/builddir/build/BUILDROOT/python-pytest-qt-4.4.0-2.fc41.x86_64/usr/lib/python3.12/site-packages/pytestqt/wait_signal.py", line 219, in _quit_loop_by_signal
    self._cleanup()
  File "/builddir/build/BUILDROOT/python-pytest-qt-4.4.0-2.fc41.x86_64/usr/lib/python3.12/site-packages/pytestqt/wait_signal.py", line 226, in _cleanup
    _silent_disconnect(signal, self._quit_loop_by_signal)
  File "/builddir/build/BUILDROOT/python-pytest-qt-4.4.0-2.fc41.x86_64/usr/lib/python3.12/site-packages/pytestqt/wait_signal.py", line 741, in _silent_disconnect
    signal.disconnect(slot)
SystemError: <class 'RuntimeError'> returned a result with an exception set
________________________________________________________________________________
----------------------------- Captured stderr call -----------------------------
Exceptions caught in Qt event loop:
________________________________________________________________________________
RuntimeError: Internal C++ object (Obj) already deleted.
The above exception was the direct cause of the following exception:
[...]

PR #596, opened 2025-03-22
Fix and clean up timer handling in signal/callback blockers

The-Compiler:
- Always create the timeout `QTimer`, even if it's not going to be used.
- Connect the signal once in `_AbstractSignalBlocker`/`CallbackBlocker.__init__` and never disconnect it.
- When checking whether a timeout was set, simply check `self._timeout` instead of `self._timer`.

Before this change, we conditionally created a `QTimer` and deleted it again when cleaning up. In addition to coming with hard to follow complexity, this also caused multiple issues around this timer:

Warnings about disconnecting signal
===================================

In `AbstractSignalBlocker._cleanup`, we attempted to disconnect `self._timer.timeout()` from `self._quit_loop_by_timeout`, under the condition that a timer was created in `__init__` (i.e. `self.timeout` was not `0` or `None`).

However, the signal only got connected in `AbstractSignalBlocker.wait()`. Thus, if `AbstractSignalBlocker` is used as a context manager with a timeout and the signal is emitted inside it:

- `self._timer` is present
- The signal calls `self._quit_loop_by_signal()`
- `self._cleanup()` gets called from there
- That tries to disconnect `self._timer.timeout()` from `self._quit_loop_by_timeout()`
- Since `self.wait()` was never called, the signal was never connected
- Which then results in either an exception (PyQt), an internal SystemError (older PySide6) or a warning (newer PySide6).

In 560f565c062abb60fbba44719649e50e56f49cca and later 81b317c5e2170b135076958a730ef62bdb448317 this was fixed by ignoring `TypeError`/`RuntimeError`, but that turned out to not be sufficient for newer PySide versions: #552, #558.

The proper fix for this is to not attempt to disconnect a signal that was never connected, which makes all the PySide6 warnings go away (and thus we can run our testsuite with `-Werror` now).

As a drive-by fix, this also removes the old `filterwarnings` mark definition, which was introduced in 95fee8b20ccd20cde728dc4e05a488797fd19b93 (perhaps as a stop-gap for older pytest versions?) but shouldn't be needed anymore (and is unused since e33880994f85bdb336f943764ffb6ad0dacc366e).

# AttributeError when a signal/callback is emitted from a different thread 

If a signal is emitted from a thread, `_AbstractSignalBlocker._cleanup()` also gets called from the non-main thread (Qt will complain: "QObject::killTimer: Timers cannot be stopped from another thread").

If the signal emission just happened to happen between the None-check and calling `.start()` in `wait()`, it failed with an `AttributeError`:

Main thread in `AbstractSignalBlocker.wait()`:

```python
if self._timer is not None:  # <--- this was true...
    self._timer.timeout.connect(self._quit_loop_by_timeout)
    # <--- ...now here, but self._timer is None now!
    self._timer.start()
````

Emitting thread in `AbstractSignalBlocker._cleanup()` via `._quit_loop_by_signal()`:

```python
if self._timer is not None:
    ...
    self._timer = None  # <--- here
```

In SignalBlocker.connect, we used:

```python
actual_signal.connect(self._quit_loop_by_signal)
```

which by default is supposed to use a `QueuedConnection` if the signal gets emitted in a different thread:

    https://doc.qt.io/qt-6/qt.html#ConnectionType-enum
    (Default) If the receiver lives in the thread that emits the signal,
    `Qt::DirectConnection` is used. Otherwise, `Qt::QueuedConnection` is used. The
    connection type is determined when the signal is emitted.

though then that page says:

    https://doc.qt.io/qt-6/qobject.html#thread-affinity
    Note: If a `QObject` has no thread affinity (that is, if `thread()` returns
    zero), or if it lives in a thread that has no running event loop, then it
    cannot receive queued signals or posted events.

Which means `AbstractSignalBlocker` needs to be a `QObject` for this to work.

However, that means we need to use `qt_api.QtCore.QObject` as subclass, i.e. at import time of `wait_signal.py`. Yet, `qt_api` only gets initialized in `pytest_configure` so that it's configurable via a pytest setting.

Unfortunately, doing that is tricky, as it means we can't import `wait_signal.py` at all during import time, and people might do so for e.g. type annotations.

With this refactoring, the `AttributeError` is out of the way, though there are other subtle failures with multi-threaded signals now:

1) `_quit_loop_by_signal()` -> `_cleanup()` now simply calls `self._timer.stop()` without setting `self._timer` to `None`. This still results in the same Qt message quoted above (after all, the timer still doesn't belong to the calling thread!), but it looks like the test terminates without any timeout anyways. From what I can gather, while the "low level timer" continues to run (and waste a minimal amount of resources), the QTimer still "detaches" from it and stops running. The commit adds a test to catch this case (currently marked as xfail).

2) The main thread in `wait()` can now still call `self._timer.start()` without an `AttributeError`. However, in theory this could restart the timer after it was already stopped by the signal emission, with a race between `_cleanup()` and `wait()`.

See #586. This fixes the test-case posted by the reporter (added to the testsuite in a simplified version), but not the additional considerations above.

The same fix is also applied to `CallbackBlocker`, though the test there is way more unreliable in triggering the issue, and thus is skipped for taking too long.

### DIFF
--- a/src/pytestqt/wait_signal.py
+++ b/src/pytestqt/wait_signal.py
@@ -24,12 +24,12 @@ class _AbstractSignalBlocker:
         self.raising = raising
         self._signals = None  # will be initialized by inheriting implementations
         self._timeout_message = ""
-        if timeout is None or timeout == 0:
-            self._timer = None
-        else:
-            self._timer = qt_api.QtCore.QTimer(self._loop)
-            self._timer.setSingleShot(True)
+
+        self._timer = qt_api.QtCore.QTimer(self._loop)
+        self._timer.setSingleShot(True)
+        if timeout is not None:
             self._timer.setInterval(timeout)
+        self._timer.timeout.connect(self._quit_loop_by_timeout)
 
     def wait(self):
         """
@@ -43,11 +43,13 @@ class _AbstractSignalBlocker:
             return
         if self.timeout is None and not self._signals:
             raise ValueError("No signals or timeout specified.")
-        if self._timer is not None:
-            self._timer.timeout.connect(self._quit_loop_by_timeout)
-            self._timer.start()
 
         if self.timeout != 0:
+            if self.timeout is not None:
+                # asserts as a stop-gap for possible multithreading issues
+                assert not self.signal_triggered
+                self._timer.start()
+                assert not self.signal_triggered
             qt_api.exec(self._loop)
 
         if not self.signal_triggered and self.raising:
@@ -62,10 +64,7 @@ class _AbstractSignalBlocker:
     def _cleanup(self):
         # store timeout message before the data to construct it is lost
         self._timeout_message = self._get_timeout_error_message()
-        if self._timer is not None:
-            _silent_disconnect(self._timer.timeout, self._quit_loop_by_timeout)
-            self._timer.stop()
-            self._timer = None
+        self._timer.stop()
 
     def _get_timeout_error_message(self):
         """Subclasses have to implement this, returning an appropriate error message for a TimeoutError."""
@@ -649,12 +648,12 @@ class CallbackBlocker:
         self.kwargs = None
         self.called = False
         self._loop = qt_api.QtCore.QEventLoop()
-        if timeout is None:
-            self._timer = None
-        else:
-            self._timer = qt_api.QtCore.QTimer(self._loop)
-            self._timer.setSingleShot(True)
+
+        self._timer = qt_api.QtCore.QTimer(self._loop)
+        self._timer.setSingleShot(True)
+        if timeout is not None:
             self._timer.setInterval(timeout)
+        self._timer.timeout.connect(self._quit_loop_by_timeout)
 
     def wait(self):
         """
@@ -664,8 +663,7 @@ class CallbackBlocker:
         __tracebackhide__ = True
         if self.called:
             return
-        if self._timer is not None:
-            self._timer.timeout.connect(self._quit_loop_by_timeout)
+        if self.timeout is not None:
             self._timer.start()
         qt_api.exec(self._loop)
         if not self.called and self.raising:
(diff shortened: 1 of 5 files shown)


The-Compiler:
Ran into some unexpected issues on CI, but now all should be fine!

src/pytestqt/wait_signal.py
```diff
@@ -43,11 +43,13 @@ def wait(self):
             return
         if self.timeout is None and not self._signals:
             raise ValueError("No signals or timeout specified.")
-        if self._timer is not None:
-            self._timer.timeout.connect(self._quit_loop_by_timeout)
-            self._timer.start()
 
         if self.timeout != 0:
+            if self.timeout is not None:
```
nicoddemus, line 48: Probably makes sense to unify these 2 ifs?

tests/conftest.py
```diff
@@ -63,10 +62,9 @@ def shutdown(self):
         def single_shot(self, signal, delay):
             t = qt_api.QtCore.QTimer(self)
             t.setSingleShot(True)
-            slot = functools.partial(self._emit, signal)
-            t.timeout.connect(slot)
+            t.timeout.connect(signal)
             t.start(delay)
-            self.timers_and_slots.append((t, slot))
+            self.timers_and_slots.append((t, signal))
```
nicoddemus, line 67: :+1:

nicoddemus (approved):
Awesome work!

src/pytestqt/wait_signal.py
nicoddemus: Probably we can drop the `if self.timeout != 0` too btw, I would trust Qt even does that already.

The-Compiler: There's a `qt_api.exec(self._loop)` below which (at least for keeping the old logic) should be run if the timeout is `None` (= infinite), but should not be run if the timeout is 0 (docs: "If the timeout is set to `0`, it’s expected that the signal arrives directly in the code inside the with `qtbot.waitSignal(...):` block.").

I noticed we *don't* have the same handling in `CallbackBlocker`, which might need a review to make sure it all makes sense and is consistent somehow, but that seems out of scope for this PR.

nicoddemus: Sure thing, thanks!

The-Compiler:
Looks like all is green and ready finally :sweat_smile: 

Should I self-merge?

nicoddemus:
> Should I self-merge?

Sure please go ahead!

PR #597, opened 2025-03-25
Test and restore compatibility with older PySide/PyQt versions

The-Compiler:
This PR:

- Reverts #593, and thus restores compatibility with PySide6 < 6.8.2, as we want to support older versions as well
- Instead makes sure that the tests run in both scenarios properly (whether `qInfo` is present or not)
- Makes sure we use the real `qInfo` (and not our `QMessageLogger` shim) if present
- Fixes exception capture tests with more fine-grained Qt backend version checks so that they also run with older PySide versions
- Actually runs the tests on CI for PyQt6 6.1 and PySide6 6.0 (both of which we claim to support according to the checks in `qt_compat.py`) to avoid future regressions
- Fixes an issue in the testsuite which leaked global patching and broke those checks

Probably best reviewed commit by commit (and not squash-merged).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,13 @@ jobs:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         qt-lib: [pyqt5, pyqt6, pyside6]
         os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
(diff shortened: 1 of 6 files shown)


nicoddemus (approved):
Thanks, appreciate the individual commits.

I fumbled on #593, thanks for fixing it!

PR #598, opened 2025-03-25
Various small MultiSignalBlocker cleanups

The-Compiler:
A couple of things I noticed while working on #594 (which I'll probably abandon as it seems to bring more trouble than it fixes).

### DIFF
--- a/src/pytestqt/wait_signal.py
+++ b/src/pytestqt/wait_signal.py
@@ -1,4 +1,6 @@
 import functools
+import dataclasses
+from typing import Any
 
 from pytestqt.exceptions import TimeoutError
 from pytestqt.qt_compat import qt_api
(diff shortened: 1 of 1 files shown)


src/pytestqt/wait_signal.py
```diff
@@ -257,12 +259,12 @@ def _get_timeout_error_message(self):
             )
 
 
+@dataclasses.dataclass
 class SignalAndArgs:
-    def __init__(self, signal_name, args):
-        self.signal_name = signal_name
-        self.args = args
+    signal_name: str
+    args: list[Any]
```
nicoddemus, line 265:
```suggestion
    args: list[object]
```

In this case, we don't care about the type of the object, for this reason `object` is safer: it will accept anything (just like `Any`), but contrary to `Any`, the type checker will flag if we accidentally try to perform some operation on it (say call a method).

The-Compiler: Yep, see https://pytest-qt.readthedocs.io/en/latest/signals.html#getting-emitted-signals-and-arguments:

> To determine which of the expected signals were emitted during a `wait()` you can use `blocker.all_signals_and_args` which contains a list of `wait_signal.SignalAndArgs` objects, indicating the signals (and their arguments) in the order they were received.

I don't think this can be made generic, as the arguments will depend on the signal being recorded, and we don't know which signals (and how many in total) there will be in there.

Good catch regarding `tuple[Any, ...]`, will adjust accordingly.

nicoddemus (approved):

PR #605, opened 2025-05-22
Type annotation infrastructure and initial typing for qtbot.py

### LINKED ISSUE #417: Type annotations
I stumbled upon this when setting a new project where I haven't configured mypy yet to just ignore pytest-qt: There are no type annotations. 

Is there a reason from this apart from “It's not there because no one did it yet”? It would be a nice-to-have thing, since the earliest mayor Python version supported is 3.6 and pytest itself has them (as far as I remember)…

herobank110:
Initial version of adding type annotations to the library.

* infrastructure of mypy in precommit (ci and local type checking)
* Typing focused around qtbot.py to start with

* Note: qt objects are typed with aliases refering to Any - this may be upgraded in future to use correct types.

Fixes #417


src/pytestqt/qtbot.py
```diff
@@ -13,6 +14,11 @@
     CallbackCalledTwiceError,
 )
 
+if TYPE_CHECKING:
+    from qtpy.QtWidgets import QWidget
```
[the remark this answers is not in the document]
herobank110: > I'm not sure we can provide useful types for Qt classes for now... as commented in https://github.com/pytest-dev/pytest-qt/issues/417#issuecomment-2901983123, I would focus on typing non-Qt things for now and getting the infrastructure in place.
> 

Would the precommit check be sufficient infrastructure? That way it runs locally as well as CI. Or shall I look into a separate CI linting stage?

src/pytestqt/qtbot.py
```diff
@@ -617,9 +668,9 @@ def captureExceptions(self):
         with capture_exceptions() as exceptions:
             yield exceptions
 
-    capture_exceptions = captureExceptions
```
herobank110, line 620: Note: removed this line becausemypy gave a warning about this also being assigned in the constructor here: https://github.com/pytest-dev/pytest-qt/blob/master/src/pytestqt/qtbot.py#L153

src/pytestqt/qtbot.py
```diff
@@ -593,7 +644,7 @@ def waitCallback(self, *, timeout=5000, raising=None):
         return blocker
 
     @contextlib.contextmanager
-    def captureExceptions(self):
+    def captureExceptions(self) -> Generator["CapturedExceptions", None, None]:
```
nicoddemus, line 647:
```suggestion
    def captureExceptions(self) -> Iterator["CapturedExceptions"]:
```

src/pytestqt/qtbot.py
```diff
@@ -725,13 +776,18 @@ def mouseRelease(*args, **kwargs):
...
 
 # provide easy access to exceptions to qtbot fixtures
-QtBot.SignalEmittedError = SignalEmittedError
-QtBot.TimeoutError = TimeoutError
-QtBot.ScreenshotError = ScreenshotError
-QtBot.CallbackCalledTwiceError = CallbackCalledTwiceError
-
-
-def _add_widget(item, widget, *, before_close_func=None):
+QtBot.SignalEmittedError = SignalEmittedError  # type: ignore[attr-defined]
```
nicoddemus, line 779: Does using `ignore[attr-defined]` here means that the type checker does not know about `QtBot.SignalEmittedError`?

If so, we should move this to the class declaration:

```python
class QtBot:
    ...
    SignalEmittedError: ClassVar[Type[SignalEmittedError]] = SignalEmittedError
```

(Same for the other exceptions of course)

.pre-commit-config.yaml
```diff
@@ -46,3 +46,7 @@ repos:
         files: ^(HOWTORELEASE.rst|README.rst)$
         language: python
         additional_dependencies: [pygments, restructuredtext_lint]
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v1.15.0'
```
nicoddemus, line 50:
```suggestion
    rev: 'v1.16.0'
```

src/pytestqt/qtbot.py
herobank110: Thanks for spotting, yes code completion seems to not see it

If they are class vars with assignment directly, type inference seems to suffice

src/pytestqt/qtbot.py
```diff
@@ -11,14 +26,31 @@
...
 
-def _parse_ini_boolean(value):
+BeforeCloseFunc = Callable[[QWidget], None]
+WaitSignalsOrder = Literal["none", "simple", "strict"]
+
+
+def _parse_ini_boolean(value: Any) -> bool:
     if value in (True, False):
-        return value
+        return cast("bool", value)
```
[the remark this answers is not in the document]
nicoddemus: Ahh thanks. 

I'm not a fan of it, but don't mind it too much either. :+1:

nicoddemus (approved):
Thanks a lot @herobank110, this looks great!

nicoddemus (approved):

nicoddemus (approved):

herobank110:
> Thanks @herobank110!
> 
> Leaving it open for a few days to give @The-Compiler a chance to review too. 👍

Keen to continue contributing with the typing and other stuff you may have

The-Compiler (changes requested):
Thank you, I'm pretty excited for this! :sparkles: 

I added a couple of comments for things I noticed, mostly they're minor things and shouldn't block anything though. A todo-list for follow-ups if we don't do them as part of this PR:

- Use [Type Hinting Generics in Standard Collections](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections) (`list`, `dict`, `type` over `typing.List` etc.)
- Use `from __future__ import annotations`
    - Use [PEP 604: New Type Union Operator](https://docs.python.org/3/whatsnew/3.10.html#pep-604-new-type-union-operator), i.e. `TheType | None` instead of `Optional[TheType]`
    - Use "bare" annotations for forward declarations instead of stringified ones
    - Configure ruff to enforce `from __future__ import annotations` on every file for consistency
- Remove redundant type annotations from Sphinx docstrings (and possibly configure Sphinx to pick up the annotations, not sure what's needed for that nowadays)

src/pytestqt/qtbot.py
```diff
@@ -723,15 +774,19 @@ def mousePress(*args, **kwargs):
     def mouseRelease(*args, **kwargs):
         qt_api.QtTest.QTest.mouseRelease(*args, **kwargs)
 
+    # provide easy access to exceptions to qtbot fixtures
+    SignalEmittedError = SignalEmittedError
+    TimeoutError = TimeoutError
+    ScreenshotError = ScreenshotError
+    CallbackCalledTwiceError = CallbackCalledTwiceError
```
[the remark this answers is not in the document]
herobank110: I believe it could introduce breaking changes to remove or alias the imports, eg if users wrote `from pytestqt.qtbot import TimeoutError`

Another option to reduce confusion in the member variables: eg `SignalEmittedError = wait_signal.SignalEmittedError`

The-Compiler (approved):

The-Compiler:
Looking great, thanks again!

nicoddemus:
No worries, it happens. :+1:

PR #648, opened 2026-05-06
Fetch signal names from PySide6 signal instances

flowln:
The signal names can be fetched by looking at the `QMetaMethod` associated with the signals. I'm not sure if PyQt also allows for that, in which case it would make sense to handle both in the same way, but at least for now this can provide some improvement.

This allows PySide6 tests to populate `signals_and_args` in `waitSignals`, which makes it compatible with PyQt. This improves the test development process by unifying handling of both versions.

### DIFF
--- a/src/pytestqt/wait_signal.py
+++ b/src/pytestqt/wait_signal.py
@@ -86,6 +86,13 @@ class _AbstractSignalBlocker:
         signal_name = signal_name.lstrip("2")
         return signal_name
 
+    def _extract_pyside_signal_name(self, potential_pyside_signal):
+        meta_method = qt_api.QtCore.QMetaMethod.fromSignal(potential_pyside_signal)
+        if not meta_method.isValid():
+            return ""
+        signal_name = meta_method.name().data().decode("utf-8")
+        return signal_name
+
     def _extract_signal_from_signal_tuple(self, potential_signal_tuple):
         if isinstance(potential_signal_tuple, tuple):
             if len(potential_signal_tuple) != 2:
@@ -117,12 +124,15 @@ class _AbstractSignalBlocker:
         signal_name = self._extract_signal_from_signal_tuple(potential_signal_tuple)
 
         if not signal_name:
-            try:
-                signal_name = self._extract_pyqt_signal_name(potential_signal_tuple)
-            except AttributeError:
-                # not a PyQt signal
-                # -> no signal name could be determined
-                signal_name = ""
+            if qt_api.is_pyqt:
+                try:
+                    signal_name = self._extract_pyqt_signal_name(potential_signal_tuple)
+                except AttributeError:
+                    # not a PyQt signal
+                    # -> no signal name could be determined
+                    signal_name = ""
+            elif qt_api.is_pyside:
+                signal_name = self._extract_pyside_signal_name(potential_signal_tuple)
 
         return signal_name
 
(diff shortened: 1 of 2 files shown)


tests/test_wait_signal.py
```diff
@@ -905,7 +905,7 @@ def test_empty_when_no_signal(self, qtbot, signaller):
             pass
         assert blocker.all_signals_and_args == []
 
-    def test_empty_when_no_signal_name_available(self, qtbot, signaller):
+    def test_non_empty_on_pyside(self, qtbot, signaller):
```
The-Compiler, line 908: The docstring and the `pytest.skip` call below don't make sense anymore with this change.

However, if we can adjust `get_mixed_signals_with_guaranteed_name()` to not require special handling of PySide anymore, I think the tests could be simplified a lot (no need for that helper function anymore), and then e.g. this test is probably completely superfluous anyways?

tests/test_wait_signal.py
```diff
@@ -1188,16 +1188,13 @@ def test_strict_order_violation(self, qtbot, signaller):
             "Missing: [signal(), signal_args(QString,int), signal_args(QString,int)]"
         ).format(signal_args, signal_args)
 
-    def test_degenerate_error_msg(self, qtbot, signaller):
+    def test_error_msg_on_pyside(self, qtbot, signaller):
```
The-Compiler, line 1191: As above, when adjusting (or removing) `get_mixed_signals_with_guaranteed_name`, I think this won't be needed as a separate test anymore.

src/pytestqt/wait_signal.py
```diff
@@ -86,6 +86,13 @@ def _extract_pyqt_signal_name(self, potential_pyqt_signal):
         signal_name = signal_name.lstrip("2")
         return signal_name
 
+    def _extract_pyside_signal_name(self, potential_pyside_signal):
+        meta_method = qt_api.QtCore.QMetaMethod.fromSignal(potential_pyside_signal)
```
The-Compiler, line 90: FWIW, on PyQt (both 5 and 6), I get `AttributeError: type object 'QMetaMethod' has no attribute 'fromSignal'`, so it looks like the split between PyQt/PySide is still needed.

The-Compiler (changes requested):
Oh, that's really nice! I think the tests will need some clean up, but really glad to hear that this actually works, I had no idea!

The-Compiler:
Oh, looking at the CI results, PySide 6.0 also lacks `QMetaMethod.fromSignal`. So we will either need to special-case PySide 6.0 and fall back to the previous logic there (including keeping those tests around I suppose), or we start requiring PySide >= 6.1 (given that 6.0 has been EOL upstream ever since 6.1 got released, almost exactly 5 years ago).

@nicoddemus what do you think?

flowln:
> Oh, looking at the CI results, PySide 6.0 also lacks `QMetaMethod.fromSignal`. So we will either need to special-case PySide 6.0 and fall back to the previous logic there (including keeping those tests around I suppose), or we start requiring PySide >= 6.1 (given that 6.0 has been EOL upstream ever since 6.1 got released, almost exactly 5 years ago).

It's amazing that the CI caught that! Great idea to run jobs with the earliest supported version!

It seems like official support for the `SignalInstance`-related functionality dates back to 6.2.0:
image

PR #649, opened 2026-05-06
Cleanup timer in signal blocker from the correct thread

flowln:
According to the [PySide6 documentation on QTimer](https://doc.qt.io/qtforpython-6/PySide6/QtCore/QTimer.html#detailed-description):

> Qt uses the timer’s thread affinity to determine which thread will emit the timeout() signal.
Because of this, you must start and stop the timer in its thread;

Because of that, we need to schedule the stop invokation to run in the timer's parent thread. This ensures correct behavior when handling signals emitted from different threads.

### DIFF
--- a/src/pytestqt/wait_signal.py
+++ b/src/pytestqt/wait_signal.py
@@ -69,7 +69,17 @@ class _AbstractSignalBlocker:
     def _cleanup(self):
         # store timeout message before the data to construct it is lost
         self._timeout_message = self._get_timeout_error_message()
-        self._timer.stop()
+
+        timer_meta = self._timer.metaObject()
+        timer_stop_signature = (
+            timer_meta.normalizedSignature("stop").data().decode("utf-8")
+        )
+        timer_stop_method = timer_meta.method(
+            timer_meta.indexOfMethod(timer_stop_signature)
+        )
+        timer_stop_method.invoke(
+            self._timer, qt_api.QtCore.Qt.ConnectionType.QueuedConnection
+        )
 
     def _get_timeout_error_message(self):
         """Subclasses have to implement this, returning an appropriate error message for a TimeoutError."""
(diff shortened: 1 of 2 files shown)


src/pytestqt/wait_signal.py
```diff
@@ -69,7 +69,17 @@ def _quit_loop_by_timeout(self):
...
+        timer_meta = self._timer.metaObject()
+        timer_stop_signature = (
+            timer_meta.normalizedSignature("stop").data().decode("utf-8")
+        )
+        timer_stop_method = timer_meta.method(
+            timer_meta.indexOfMethod(timer_stop_signature)
+        )
+        timer_stop_method.invoke(
+            self._timer, qt_api.QtCore.Qt.ConnectionType.QueuedConnection
+        )
```
The-Compiler, line 82: I think this can be simplified to:

```suggestion
        assert qt_api.QtCore.QMetaObject.invokeMethod(self._timer, "stop", qt_api.QtCore.Qt.ConnectionType.QueuedConnection)
```

For some reason I don't quite understand, your approach seems to fail on PyQt, while this seems to work fine there too.

I think it's a good idea to add an `assert` for the return value, since calling this should never fail. If it does anyways, I think that should be a loud and not a silent failure. I'm surprised that PyQt actually does turn the failure into a `RuntimeError` on its own, I don't think that's guaranteed to happen everywhere though.

The-Compiler (changes requested):
Good catch! I'm not sure why, but this seems to fail on PyQt with a non-descriptive error message. I found a simpler way to do (hopefully) the same thing though, and that seems to work there too.

src/pytestqt/wait_signal.py
flowln: Oh yeah your approach is way cleaner! I was afraid of the `stop` string not being sufficient for the `invokeMethod` call, since sometimes those methods require the full signature string (method index + method signature). Glad that's not the case here!

